PR #647, opened 2019-04-05
Parserx

Skillmon:
I've added a `parserx` module which is similar to the `parser` module but extends it such that one can define actions taking arguments. There are some other additions as well. The two can coexist, `parserx` doesn't reuse any of the internals of `parser`. I've added a short documentation as well. The module works in plain and LaTeX. I didn't test Context. I built the pgfmanual both with pdftex and luatex.


hmenke:
I haven't reviewed this in detail yet, but there are two issues with `\protected`.

1. Even though it is a primitive, it is format-dependent because ConTeXt renames `\protected` to `\unexpanded` (or `\normalprotected`).  Format differences are handled in [`pgfutil-context.def`](https://github.com/pgf-tikz/pgf/blob/master/tex/generic/pgf/utilities/pgfutil-context.def), [`pgfutil-latex.def`](https://github.com/pgf-tikz/pgf/blob/master/tex/generic/pgf/utilities/pgfutil-latex.def), and [`pgfutil-plain.def`](https://github.com/pgf-tikz/pgf/blob/master/tex/generic/pgf/utilities/pgfutil-plain.def).
2. `\protected` requires e-TeX, which for now I don't want to opt-in to, because people have been complaining (see #507).

It also looks like `\parserxdef` is just `\parserdef` with an additional optional argument, so it would make sense to just extend the existing parer module rather than introducing another slightly different one.

davidcarlisle:
@hmenke  without commenting on the specifics of this PR, is a no-etex policy really sustainable?


```
\input tikz


z
\bye
```

fails already with 

````
! Undefined control sequence.
\pgfkeyssetevalue ...gfkeys@temptoks =\scantokens
````

(Same error if you just input `pgf`)

**Comment by hmenke** [comment]
@davidcarlisle Can't reproduce on latest master.

**Comment by davidcarlisle** [comment]
@hmenke  ah sorry, that quick test was from the tl2019 public version but still it indicates that avoiding etex is going to be increasingly hard (with almost no benefit)  given that it is required by latex and context and built in to pdftex, luatex and xetex, so you are aiming at plain tex users who can't use etex, which is just a handful of users most likely.

**Comment by hmenke** [comment]
@davidcarlisle I take a different position.  As long as TeX Live ships Knuth TeX as the `tex` binary, people will try to typeset PGF code with it and send me bug reports about broken plain TeX compatibility.  Instead TeX Live should ship Knuth TeX as a `tex90` binary (or so) and e-TeX as the `tex` binary.

**Comment by hmenke** [comment]
@Skillmon Another stylistic problem in your code are the `<<<` and `>>>` comments.  Chevrons are already used by Git to indicate merge conflicts (see the [`git-merge` documentation](https://git-scm.com/docs/git-merge#_true_merge)).  So these unfortunately have to go.

**Comment by hmenke** [comment]
@Skillmon I have just committed the wrapper `\pgfutil@protected` for `\protected` which will do nothing if `\protected` is not available and otherwise choose the appropriate version of `\protected` given by the format.

**Comment by Skillmon** [comment]
@hmenke will edit. It also uses `\numexpr`, will edit that one out, too. I did not build it into the `parser` module, because it changes too much.

Also on the `>>>` and `<<<` issue: Those are the fold marks of my VIM. Git merge uses way more chevrons (5-8 or so) so I never had problems with these. But if you want me to, I can remove those.

**EDIT:** just noticed that you're right, git also uses `>>>` (I just never saw those before, tbh.). Now I need a new fold mark for my VIM (one that doesn't conflict with git and that isn't build up by braces, because three consecutive braces are too common in TeX codes).

**Comment by Skillmon** [comment]
My (really small) tests work now for `tex`, `pdftex` and `pdflatex`. I also removed the foldmarks. Also replaced some unnecessary new definitions with `\pgfutil@...` variants.

**Comment by hmenke** [comment]
Do you see a possibility to integrate this into the existing parser module?
If you really have to make this a separate module, please choose another name.  `parserx` is not very descriptive.  I also don't particularly like the new trend in L3 of adding `x` to every package name (`xparse`, `xgalley`, `xcoffin`, `xfrac`, etc.).

**Comment by Skillmon** [comment]
I gave it a try and the pgfmanual builds (tried the LuaTeX version) with `parser` replaced by my module. I don't know whether this has the potential of breaking anything else to be honest. There currently is the (minor) incompatibility that by default blank space is defined as a no-op in my module, I could change that, if you think it should be changed.

So should I replace the `parser` module?

**Comment by hmenke** [comment]
Currently the parser is only used in
https://github.com/pgf-tikz/pgf/blob/81bac3c5643005a3a3a89d490c35ecb19554c941/tex/generic/pgf/libraries/pgflibrarysvg.path.code.tex#L13
If the manual still builds with `parserx` and the SVG example doesn't break, I would suggest you merge your `pgfmoduleparserx.code.tex` into the existing `pgfmoduleparser.code.tex` so that the diff doesn't look to horrible.

Maybe your new parser might even fix #628.

**Comment by Skillmon** [comment]
I doubt that it can fix that without further code. All it does (or was meant to do) is add a mechanism for arguments into a letter-by-letter parser. However I wrote an expandable test that checks whether something is a valid float expression for something else. If you want me to, I can take a look into the SVG parser and perhaps add that test.

The documentation page of the `pgf.path` library looks the same with my parser replacing the `parser` module, so I think we could replace it (though I'll have to rewrite some documentation for that). Hopefully I will push later today.

**Comment by Skillmon** [comment]
What about the initial definition of spaces being a no-op? Should I remove that to be as compatible to the old module as possible?

**Comment by Skillmon** [comment]
Ok, now parser is completely replaced by my module and the documentation of the new `parser` module should be fine (a proofread might be a good idea though). I've yet to look into [#628](https://github.com/pgf-tikz/pgf/issues/628), but I don't think my parser alone can fix that. Instead it should be fixable even with the old parser (if it is done how I think it is, as I said, I didn't look into it yet).

File: tex/generic/pgf/libraries/pgflibrarysvg.path.code.tex [hunk 1]
```diff
@@ -12,6 +12,9 @@
 
 \usepgfmodule{parser}%
 
+% pgfmath is needed to parse scientific notation
+\input pgfmath.code.tex
```
hmenke: This should already be available at this point.  It is loaded in
https://github.com/pgf-tikz/pgf/blob/f9d662d09de0d5613a0392c9986dece22dd3f913/tex/generic/pgf/basiclayer/pgfcore.code.tex#L15

doc/generic/pgf/text-en/pgfmanual-en-module-parser.tex
```diff
@@ -57,35 +60,222 @@ \section{Parser Module}
...
-
-    When you set the \meta{state} to |all|, the state \meta{action} is
-    performed in all states as a fallback, whenever \meta{symbol meaning} is
-    encountered. This means that when you do not specify anything explicitly
-    for a state and a letter, but you do specify something for |all| and this
-    letter, then the specified \meta{action} will be used.
-
-    When the parser encounters a letter for which nothing is specified in the
-    current state (neither directly nor indirectly via |all|), an error occurs.
+\begin{command}
```
hmenke:
```latex
\begin{command}%
```
TeX might not be in state S (skipping spaces) at this point.

tex/generic/pgf/modules/pgfmoduleparser.code.tex
```diff
@@ -11,101 +11,670 @@
...
+          {%
+            \pgfparser@getnexttoken@iftokengobble\pgfutil@sptoken{ }%
+              {\pgfparser@getnexttoken@defchar}%
+          }%
+      }%
+  }
+\pgfutil@protected\def\pgfparser@getnexttoken@iftokengobble#1#2%
+  {%
+    \ifx\pgfparsertoken#1%
+      \def\pgfparserletter{#2}%
```
hmenke, line 517: Isn't `#2` always a single token?  In that case you could use
```latex
\let\pgfparserletter=#2%
```
That is actually very important because an `\ifx` check might fail if `#2` is a `\long` macro, e.g.
```latex
\long\def\foo{abc}
\def\pgfparserletter{\foo}
%\let\pgfparserletter=\foo
\ifx\pgfparserletter\foo
  Yes
\else
  No
\fi
```
With `\def` it will print “No”, but with `\let` it will print “Yes”.

tex/generic/pgf/modules/pgfmoduleparser.code.tex
```diff
@@ -11,101 +11,670 @@
...
+  }
+\pgfutil@protected\def\pgfparser@getnexttoken@gobble\fi#1#2%
+  {%
+    \fi
+    \afterassignment\pgfparser@getnexttoken@b
+    \let\pgfparser@ignored@token= % gobble should gobble a space
+  }
+\pgfutil@protected\long\def\pgfparser@getnexttoken@defchar#1%
+  {%
+    \def\pgfparserletter{#1}%
```
hmenke, line 530: Same as for the other `\def\pgfparserletter`.

tex/generic/pgf/modules/pgfmoduleparser.code.tex
```diff
@@ -1,4 +1,4 @@
-% Copyright 2009 by Till Tantau
```
hmenke: Don't remove copyright statements of other people.

hmenke (changes requested):

tex/generic/pgf/modules/pgfmoduleparser.code.tex
Skillmon: That is done because `\pgfparsertoken` is already `\let` to what ever follows, `\pgfparserletter` isn't needed by the module itself, but it might be useful for someone who writes a parser in the `\pgfparserdefunknown` code (because getting back a token from its `\meaning` is very hard so if one needs the actual letter/token this is convenient). Also `#2` might be a single space (there are 3 possibilities: `\bgroup`, `\egroup` and ` ` (space))

tex/generic/pgf/modules/pgfmoduleparser.code.tex
Skillmon: Same answer as for the other `\def\pgfparserletter`.

tex/generic/pgf/modules/pgfmoduleparser.code.tex
Skillmon: I removed it because it doesn't share a single line (and if so that is by accident). In the documentation Till's copyright isn't removed because I used almost all of his text. If removing it under these conditions was false, I apologize, it wasn't my intention to steal any interlectual property.

tex/generic/pgf/libraries/pgflibrarysvg.path.code.tex
Skillmon: It does no harm inputting it twice `pgfmath.code.tex` has safety measures against that, and I wasn't sure whether it will be loaded in every case this library might be loaded.

tex/generic/pgf/modules/pgfmoduleparser.code.tex
hmenke: Your parser is based on Till's original version, making it a derived work.  The GPL requires that derived works contain the same copyright notices as the original.

But apart from the legal aspect, I think Till deserves some credit.  Your parser would probably not exist without his version.

tex/generic/pgf/modules/pgfmoduleparser.code.tex
Skillmon: It is explained in the documentation.

doc/generic/pgf/text-en/pgfmanual-en-module-parser.tex
Skillmon: As far as I understood, `command` takes an argument so spaces before that should be ignored.

Skillmon:
Every requested change incorporated except for the `\pgfparserletter` changes, additionally commented many uncommented line endings in `pgflibrarysvg.path.code.tex` (though those most likely did no harm, as PGF ignores those anyway, as far as I know).

tex/generic/pgf/modules/pgfmoduleparser.code.tex
Skillmon: Yes, that is correct. Though my parser is really only based on the documentation in the manual, not on his code. Still your point is valid.

hmenke (approved):

PR #653, opened 2019-04-09
- minor improvement

Mo-Gul:
Just a test if I have found out how it works.


hmenke (commented):
What's the improvement?

hmenke:
Yes, that works :)

Mo-Gul:
I try to avoid line breaks in commands, because some of them lead to multiple spaces in the PDF (because of the indention). I guess that this is not the case for this command, but also I try to break the line after 80 chars, which is true here. But of course this is just a (very) minor improvement ... :)

Issue #658, opened 2019-04-17
Mandelbrot set shading fails with dvips driver


dcpurton:
The Mandelbrot set shading from the Shadings library does not work with the dvips driver.

Consider:

```TeX
\documentclass{article}
\usepackage{tikz}
\usetikzlibrary{shadings}
\begin{document}
\tikz \shade[shading=Mandelbrot set] (0,0) rectangle (2,2);
\end{document}
```

It gives GhostScript errors at the `ps2pdf` step:

```
Error: /undefinedresult in --mul--
Operand stack:
   --nostringval--   --nostringval--   --nostringval--   --nostringval--   -13.0006   -2.00064   -1.00064   -0.000640869   0.999359   1.99936   2.99936   3.99936   -12.0006   -1.00064   -0.000640869   0.999359   1.99936   2.99936   3.99936   -11.0006   -3.00064   -2.00064   -1.00064   -0.000640869   0.999359   1.99936   2.99936   3.99936   -10.0006   -6.00064   -10.0006   -0.480051   -0.800051   -5.01942e+21   4.82998e+20   4.82998e+20
Execution stack:
   %interp_exit   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--   false   1   %stopped_push   1992   1   3   %oparray_pop   1991   1   3   %oparray_pop   1979   1   3   %oparray_pop   1833   1   3   %oparray_pop   --nostringval--   %errorexec_pop   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   -9.00064   1.0   50.0006   --nostringval--   %for_real_continue   --nostringval--   -5.00064   1.0   50.0006   --nostringval--   %for_real_continue   --nostringval--   --nostringval--
Dictionary stack:
   --dict:724/1123(ro)(G)--   --dict:1/20(G)--   --dict:85/200(L)--   --dict:178/300(L)--   --dict:61/200(L)--   --dict:61/200(L)--
Current allocation mode is local
Current file position is 31610
GPL Ghostscript 9.27: Unrecoverable error, exit code 1
```

It doesn't produce errors with 9 iterations, but the output is wrong.

This is the only functional shading in the Shadings library that fails.

hmenke:
It didn't even work in c28bd269837dba9f5b9e5b0543846646a0b70d6b which is where it was introduced 11 years ago.

[...]

PR #661, opened 2019-04-20
Fix Mandelbrot set shading definition #658

dcpurton:
Fix PostScript code in Mandelbrot set functional shading definition.


hmenke (approved):

hmenke:
Thank you very much for your contribution.

PR #664, opened 2019-04-24
Implement shadings in CMYK and grayscale (resolves #553)

### LINKED ISSUE #553: Shade does not respect color space
*Migrated from SourceForge*
Author: sauert
Timestamp: 2007-12-17 08:30:05

Hello,

shaded paths are always rendered in RGB even though all three colors are given in CMYK. Filled path on the other hand are correctly rendered in the color model of the given color. This results in color "breaks" if a shaded area adjoins a filled area of the same color. 
As fas as I understand the PDF specs the color space for shadings can also be chosen freely.

I attached a minimal example compiled with pdflatex.

Yours Bastian

dcpurton:
I think I've tested all that I can think of for now with this.

Apart from issues that currently affect the existing shading implementation I'm happy with how it works.

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-base-shadings.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-base-shadings.tex
@@ -47,12 +47,86 @@ rgb(0cm)=(1,0,0); rgb(2cm)=(0,1,0); rgb(4cm)=(0,0,1)
 This line means that at 0cm (the left end) of the bar, the color should be red,
 which has red-green-blue (rgb) components (1,0,0). At 2cm, the bar should be
 green, and at 4cm it should be blue. Instead of |rgb|, you can currently also
-specify |gray| as color model, in which case only one value is needed, or
+specify |cmyk| as color model, in which case four values are needed,
+|gray| as color model, in which case only one value is needed, or
 |color|, in which case you must provide the name of a color in parentheses. In
 a color specification the individual specifications must be separated using a
 semicolon, which may be followed by a whitespace (like a space or a newline).
 Individual specifications must be given in increasing order.
 
+\subsubsection{Color models}
+
+An attempt is made to produce shadings consistent with the currently selected
+|xcolor| package color model. The |rgb|, |cmyk|, and |gray| color models from
+the |xcolor| package are supported.
+
+\textbf{Note:} The color model chosen for a shading is based on the |xcolor|
+color model \emph{at the time the shading is created}. This is either when
+\cs{pgfdeclare*shading} is called with no optional argument or when
+\cs{pgfuseshading} is called if \cs{pgfdeclare*shading} was called with an
+optional argument.
+
+If the |xcolor| package |natural| color model is in use then the shading color
+model will be \textsc{rgb} by default. In practice this means that if you are
+using the |natural| color model of the |xcolor| package you can get mismatched
+colors if you, for example, create a shading from green (which is defined as
+\textsc{rgb}) to magenta (which is defined as \textsc{cmyk}). The shading will
+finish with \textsc{rgb} magenta which will look different to the
+\textsc{cmyk} magenta used in solid colors.
+
+You can avoid mismatched colors by loading the |xcolor| package first with an
+explicit color model (|rgb|, |cmyk|, or |gray|).
+
+\begin{codeexample}[code only]
+\begin{tikzpicture}
+  \fill[green] (0,0) rectangle (1,1);
+  \shade[left color=green, right color=magenta] (1.25,0) rectangle (3.75,1);
+  \fill[magenta] (4,0) rectangle (5,1);
+\end{tikzpicture}
+\end{codeexample}
+
+\begin{center}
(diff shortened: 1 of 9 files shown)


doc/generic/pgf/text-en/pgfmanual-en-base-shadings.tex
```diff
@@ -47,12 +47,86 @@ \subsection{Overview}
...
-specify |gray| as color model, in which case only one value is needed, or
+specify |cmyk| as color model, in which case four values are needed,
+|gray| as color model, in which case only one value is needed, or
 |color|, in which case you must provide the name of a color in parentheses. In
 a color specification the individual specifications must be separated using a
 semicolon, which may be followed by a whitespace (like a space or a newline).
 Individual specifications must be given in increasing order.
 
+\subsubsection{Color models}
+
```
hmenke, line 58: I hope I can review this tomorrow.  In the meantime, please add your copyright to all the files and also a short note
```latex
\subsubsection{Color models}

\noindent\emph{by David Purton}
```
in the section in the manual (if you want).  I think you deserve credit for this.

hmenke (changes requested):

doc/generic/pgf/text-en/pgfmanual-en-base-shadings.tex
dcpurton: I'll add something like:

```
% CMYK and grayscale shadings adaptation copyright 2019 by David Purton
```

to all the files (except `pgfsys.code.tex`). Is that the kind of thing you have in mind? Though the bulk of the changes are really only found in `pgfcoreshade.code.tex`.

dcpurton:
Actually, Commit 492f13c got me thinking. I think maybe this and 4257d89 are not actually necessary. I had something like this in `pgf-cmykshadings`, but now I'm not convinced.

I think it's enough to do this in `\pgfuseshading`:

```Patch
- \expandafter\pgfutil@convertcolorspec\pgf@tempcolor{rgb}{\pgf@rgbcolor}%
- \edef\pgf@shade@adds{\pgf@shade@adds,\pgf@rgbcolor}%
+ \expandafter\pgfutil@convertcolorspec\pgf@tempcolor{\pgf@shading@model}{\pgf@color}}%
+ \edef\pgf@shade@adds{\pgf@shade@adds,\pgf@color}%
```

What's the best way to fix this now?

Does the pull request break if I remove these two commits, add in the above changes (they probably belong with 3f125fe) then rebase? Or do I have to add revert commits and a new one for the above changes? Just thinking about keeping the commit history tidy for this feature.

hmenke:
You can rebase, squash, and merge as you please.  When you force-push to your own branch, GitHub will automatically refresh this pull request.

doc/generic/pgf/text-en/pgfmanual-en-base-shadings.tex
```diff
@@ -313,9 +395,184 @@ \subsubsection{General (Functional) Shadings}
...
+  70 sub dup mul add sqrt
+  dup mul neg 1.002 exch exp 1.0 exch sub
+  1.0 3 1 roll
+  \ifpgfshadingmodelcmyk
+    \pgffuncshadingrgbtocmyk
+  \fi
+  \ifpgfshadingmodelgray
+    \pgffuncshadingrgbtogray
+  \fi
+}
```
hmenke, line 496: Do you think it would make sense to just automatically append the bit
```latex
  \ifpgfshadingmodelcmyk
    \pgffuncshadingrgbtocmyk
  \fi
  \ifpgfshadingmodelgray
    \pgffuncshadingrgbtogray
  \fi
```
to `\pgfdeclarefunctionalshading`?

hmenke (commented):
Just this one question.  Otherwise I'm very happy with your contribution.  This is a great new feature for PGF!

doc/generic/pgf/text-en/pgfmanual-en-base-shadings.tex
dcpurton: Yes, this could make sense and I'm happy to adjust it, but there would need to be a package option (or some other flag) to enable or disable automatic conversion.

The disadvantage of making it automatic (with no option) is that it forces you to always work in RGB. But the conversion from RGB to CMYK makes assumptions regarding under colour removal and black generation that you might not want if you were working in CMYK directly. Since there are no colour profiles involved, I think it is important to be able to work in your intended output colour space directly.

So a few of questions:

1. Do you have a preference for how this option would be controlled? (e.g., a package option, a Boolean somewhere, something else, etc.)
2. Could `\ifpgfshadingmodel*` and `\pgffuncshadingrgbto*` then become private and removed from the manual?
2. Are the `\pgffuncshadingcmykto*` and `\pgffuncshadinggrayto*` macros needed at all?

hmenke: I think this is fine for now and I will just merge it.  Before making a new release we can think about it again.  I don't think `\pgfdeclarefunctionalshading` is widely used.

hmenke (approved):

PR #668, opened 2019-04-29
corrected some typos and re-formatted newly added stuff so that it matches the rest

Mo-Gul:


hmenke:
Please do *not* reformat existing content.  It serves absolutely no purpose and in some circumstances just breaks things, see https://github.com/pgf-tikz/pgf/commit/9cd8465397b6df6589a8d0c4d78f199533d49e4f.

Please undo the reformatting so I can merge the corrected typos.

kevinhappi:
Great

Mo-Gul:
@hmenke, sorry for the late answer, but I think reformatting *has* a purpose which is *consistency*. Don't you think so, too? In addition I don't think that I can do much harm by reformatting the *manual files*.

I (re)read all modifications to the manual so a chance of typos etc. gets minimized. By that I also format the newly added stuff to keep it consistent and new authors can (hopefully) easily identify the "formatting rules" which they (hopefully) adapt.

If you really insist that I should stop this, there is nothing more to do for me in this project because I can't do anything else than that (as I hopefully have told you a while a go).

So either you accept the merge request as is or you reject it ...

kevinhappi:
Accept

hmenke (approved):

PR #729, opened 2019-08-18
Pimp codeexamples (issue #640) part 2 (of pull request #711)

Mo-Gul:
basically this is the continuation of pull request #711, but it doesn't show up there, because I first merged branch `PimpCodeexamples` into `master`.

PR #733, opened 2019-08-27
Improved version of #729 without questionable commits

hmenke:
I rebased the complete `Mo-Gul/master` branch onto `pgf-tikz/master` and dropped the botched merges.


Mo-Gul (approved):
I cloned your fork and diffed it with my (local) fork. All my changes were applied successfully and only some stuff that is definitely not from me remained.

Many thanks for taking care!!

Issue #736, opened 2019-08-28
A bit typos in manual


XAjzh:
I am a beginner to Ti*k*Z. When I read the pgfmanual, I found a bit typos. Then I read the latest pgfmanual which is on [https://pgf-tikz.github.io/pgf/pgfmanual.pdf](https://pgf-tikz.github.io/pgf/pgfmanual.pdf) and [https://github.com/pgf-tikz/pgf/releases](https://github.com/pgf-tikz/pgf/releases). The typos still exist.

All the typos are library's name. They are listed below with bold style(copy from [https://pgf-tikz.github.io/pgf/pgfmanual.pdf](https://pgf-tikz.github.io/pgf/pgfmanual.pdf)).

part 1, section 3 *Tutorial: A Petri-Net for Hagen*, subsection 3.13 *Using Layers: The Background Rectangles*, second paragraph:

> The solution is to use layers. When the **background** library is loaded, Hagen can put parts of his picture inside a scope with the on background layer option...

part 1, section 5 *Tutorial: Diagrams as Simple Graphs*, subsection 5.4 *The Diagram as a Graph*:

> Matrices allow Ilka to align the nodes nicely, but the connections are not quite perfect. The problem is that the code does not really reflect the paths that underlie the diagram. For this, it seems natural enough to Ilka to use the **graph** library since, after all, connecting nodes by edges is exactly what happens in a graph. The **graph** library can both be used to connect nodes that have already been created, but it can also be used to create nodes “on the fly” and these processes can also be mixed.

Help this helps. Then, please don't mind my poor English.

Thank you for your work.

PR #737, opened 2019-09-03
checked documentation of `pattern.meta` library (issue #602) + other minor corrections in the manual

Mo-Gul:


Mo-Gul:
@hmenke, it would be nice if you could accept (or reject) this pull request, because then I would push another commit related to issue #736 which is a bit more complex and I don't want to mix them up here. Thanks.

hmenke:
You could just push it to a different branch.

hmenke (approved):

Mo-Gul:
Thanks for merging and thanks for the suggestion to push it to another branch. I have tried this and it seems it has worked.

PR #765, opened 2019-11-01
Add one clarifying paragraph to the tutorial.

ghost:
I spend 4 hours trying to debug  why the style "karl" work, and the style "scheme" doesn't, even made a question on tex.se.

https://tex.stackexchange.com/questions/514454/how-to-i-set-rounded-corners-inside-a-style-in-tikz/514455#514455

Got so annoyed that decided that this place may be clarified a bit. Please, don't judge my writing skills too strongly.

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
@@ -471,7 +471,7 @@ use another color.
 \end{tikzpicture}
 \end{codeexample}
 
- In this example, the definition of the style "Karl's grid" is given as an 
+ In this example, the definition of the style |Karl's grid| is given as an 
  optional argument to the |{tikzpicture}| environment. Additional styles for other
  elements would follow after a comma. With many styles in effect, the optional
  argument of the environment may easily happen to be longer than the actual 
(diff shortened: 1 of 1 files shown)


doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
```diff
@@ -471,6 +471,11 @@ \subsection{Adding a Touch of  Style}
 \end{tikzpicture}
 \end{codeexample}
 
+ In this example, the definition of the style "Karl's grid" is given as an
```
hmenke:
```suggestion
 In this example, the definition of the style ``Karl's grid'' is given as an 
```
Please use the correct quotes.  What you have typed right now will render absolutely horribly.  GitHub is even highlighting it in red.
Other than that, it looks good to me.

hmenke (changes requested):

doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
joulev: Wouldn't a monospace (code) font be better? 
```suggestion
 In this example, the definition of the style |Karl's grid| is given as an 
```

hmenke: Probably yes.

hmenke (approved):

PR #771, opened 2019-11-08
Avoid overflow errors when computing the intersection of line segments.

sasozivanovic:
Fixes the situation described in issue #369 (discussion after November 2019).

A family of `\pgfutil@ifsafeto...` macros is defined, testing whether an
operation can be safely applied to the given dimensions.
* `\pgfutil@ifsafetoadd` tests addition,
* `\pgfutil@ifsafetosubtract` tests subtraction,
* `\pgfutil@ifsafetomultiply` tests multiplication, and
* `\pgfutil@ifsafetomultiplybynumber` tests multiplication of a "floating"
  number (i.e. a unitless dimension) and a dimension; the special case is
  needed as the number can exceed \maxdimen (given in pts).

Usage: `\pgfutil@ifsafeto...{first dimen}{second dimen}{true code}{false code}`
`true code` must contain the code which performs the actual operation.  Before
`false code` is executed, `\ifpgf@dimen@overflow` is globally set to true, so
that the caller can check for overflows.

As a consequence of using `\pgfutil@ifsafeto...`, `\pgf@iflinesintersect@` now
gracefully handles large coordinates, all the way up to `\maxdimen`.  However,
when a segment is longer than `\maxdimen`, the macro still yields a false
negative.

The execution time increases by 1% - 1.5%.

### DIFF
--- a/tex/generic/pgf/libraries/pgflibraryintersections.code.tex
+++ b/tex/generic/pgf/libraries/pgflibraryintersections.code.tex
@@ -493,18 +493,31 @@
         \pgf@ya=\the\pgf@ya\space
     }%
     %
+    \let\pgf@intersect@next=\pgfutil@secondoftwo%
     % B := (2-1)
+    \pgfutil@ifsafetosubtract\pgf@xb\pgf@xa\pgf@iflinesintersect@i{}}
+\def\pgf@iflinesintersect@i{%
     \advance\pgf@xb by-\pgf@xa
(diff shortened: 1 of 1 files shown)


sasozivanovic:
I'm working on an alternative solution (still using `\pgfutil@ifsafeto...` macros) which is both more readable and not limited to segments shorter than \maxdimen.

hmenke:
It would also be nice if there was a user interface to switch back to the old behaviour.  People with enormous documents with thousands of Ti*k*Z pictures might not want to pay the additional performance penalty.

sasozivanovic:
Great idea, will do!

PR #778, opened 2019-11-13
Removing tab visible in documentation

samcarter:
The tab from the code indention within the verbatim `|[use as bounding box]|` was visible in the output. Moving the line break a bit solves the problem

Screen Shot 2019-11-12 at 22 09 33


hmenke:

PR #780, opened 2019-11-13
missing markup around |to| to distinguish it from the word "to"

samcarter:
Maybe it would make sense to write |to| instead of to?

I read the paragraph a couple of times and without this differentiation I found it very confusing to understand which tos are code and which are part of the sentance.


hmenke:

PR #781, opened 2019-11-13
docu: simplified markup

samcarter:
Is there any deeper reason for writing `|\||ifpgfrememberpicturepositiononpage|` instead of `|\ifpgfrememberpicturepositiononpage|`?

Simplifying the markup would also enable linking to the documentation of the macro.


hmenke:

PR #827, opened 2020-01-30
pgfmathparser.code.tex: add pdfTeX/LuaTeX/pTeX units

aminophen:
Attempt to add engine-specific dimension units.

* pdfTeX and LuaTeX: `nd` and `nc` are available.
* pTeX and upTeX: `zw`, `zh`, `Q` and `H` are available.

### DIFF
--- a/tex/generic/pgf/math/pgfmathparser.code.tex
+++ b/tex/generic/pgf/math/pgfmathparser.code.tex
@@ -235,6 +235,16 @@
   \pgfmath@tokens@make{dimexpr}{\dimexpr\glueexpr}
 \fi
 \pgfmath@tokens@make{unit}{{bp}{cc}{cm}{dd}{em}{ex}{in}{mm}{pc}{pt}{sp}}
+\ifx\pdftexversion\@undefined\else % for pdfTeX
+  \pgfmath@tokens@make{unit}{{nc}{nd}}
+\fi
+\ifx\directlua\@undefined\else % for LuaTeX
+  \pgfmath@tokens@make{unit}{{nc}{nd}}
+\fi
+\ifx\kanjiskip\@undefined\else % for pTeX, upTeX
+  \pgfmath@tokens@make{unit}{{H}{Q}{zh}{zw}}
+\fi
+%
 \pgfmath@tokens@make{mathunit}{{mu}}
 \pgfmath@tokens@make{numeric}{.0123456789}
 \pgfmath@tokens@make{number}{0123456789}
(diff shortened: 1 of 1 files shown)


tex/generic/pgf/math/pgfmathparser.code.tex
```diff
@@ -235,6 +235,16 @@
   \pgfmath@tokens@make{dimexpr}{\dimexpr\glueexpr}
 \fi
 \pgfmath@tokens@make{unit}{{bp}{cc}{cm}{dd}{em}{ex}{in}{mm}{pc}{pt}{sp}}
+\ifx\pdftexversion\@undefined\else % for pdfTeX
```
hmenke, line 238: Since we require e-TeX you can also use `\ifdefined`.
```suggestion
\ifdefined\pdftexversion % for pdfTeX
```
Maybe we can also use the ìftex` package.

tex/generic/pgf/math/pgfmathparser.code.tex
```diff
@@ -498,7 +511,14 @@
...
           \fi%
-        \else% ...but here it is units.
+        \else% ...but here it is single-character units.
+          \global\pgfmathunitsdeclaredtrue%
+          \pgfmath@dimen=\pgfmath@number\pgfmath@token\relax%
+          \edef\pgfmathresult{\pgfmath@tonumber{\pgfmath@dimen}}%
+          \expandafter\pgfmath@stack@push@operand\expandafter{\pgfmathresult}%
+          \let\pgfmath@parse@next=\pgfmath@parse@@operator%
+        \fi%
+        \else% ...but here it is double-character units.
```
hmenke, line 521: Maybe switch these around?  Double-character units should be processed before single-character ones.  It's not a problem right now, but imagine there were two units `m` and `mm`, then `m` would be processed, leaving another single `m` behind which is probably not the same as `mm`.

hmenke (changes requested):

tex/generic/pgf/math/pgfmathparser.code.tex
aminophen: No, the order is correct. Please look carefully at the \ifx block; first it tests double-character, then tests single-character.

tex/generic/pgf/math/pgfmathparser.code.tex
aminophen: If e-TeX is actually required, you are right, but a few lines above that line I've added says

```
\ifx\dimexpr\@undefined\else % needed for classic tex
```

so I guess this file should be intended to work also with both classix tex.

hmenke: Ah, right.  I think `pgfmath` technically does not require e-TeX yet.  However, this made me thinking whether we should just use `iftex` for engine detection.

aminophen: `iftex` started to support detecting ptex only recently (2019 by David), so it may be safer to keep my implementation as-is.

hmenke:
Rebased and merged c739cf1b

Issue #831, opened 2020-02-06
missing bracket in arrows doc


thinbold:
Version: 3.1.5b

On page 211, a right bracket is missing in this line

    >-< /.tip = >[reversed

It corresponds to this line in the doc

https://github.com/pgf-tikz/pgf/blob/master/doc/generic/pgf/text-en/pgfmanual-en-tikz-arrows.tex#L1675

hmenke:
I've committed a partial fix which at least brings back the bracket, but unfortunately it is not in italics.

PR #840, opened 2020-03-18
fix missing trailing "]" and italic opening brackets in doc

muzimuzhi:
Inside shortverb,
1. If it contains "=" and ends with "]", this "]" is missing from output.
   E.g., "|[key=val]|" produces "[key=val".
   Change the input to "|[key=val||]|".

2. If it contains "=<opening bracket>", this bracket is typeset in italic.
   E.g., "|a={...}|" produces "a=<italic {>...}".
   Change the input to "|a=||{...}|".

The first kind of change is rather complete, but for the  second kind, 
I only checked contents of every `quote` environment.

There is already a case of second kind in the source code:
https://github.com/pgf-tikz/pgf/blob/357bc05926a057f7c1fcbc95877fc508723ac105/doc/generic/pgf/text-en/pgfmanual-en-tikz-shapes.tex#L2404-L2408

---
Locally on my computer, I use a simplified `pgfmanual.tex` to test and compare:
```tex
\documentclass[a4paper]{ltxdoc}

\input{../pgfmanual-luatex.cfg}
\input{../../text-en/pgfmanual-en-main-preamble.tex}

\begin{document}
\parindent=0pt
\subsubsection*{First kind}
Before: |plot[yshift=3cm]| \par
After: |plot[yshift=3cm||]| 

\subsubsection*{Second kind}
Before
\begin{quote}
    |>-< /.tip = >[reversed]|
\end{quote}

After
\begin{quote}
    |>-< /.tip = ||>[reversed]|
\end{quote}

% pgfmanual-en-tikz-shapes.tex:2404
% (the double vertical bar after = is needed to avoid the two opening brackets
%  being typeset in italics)
Before
\begin{quote}
    |:some attribute = {|\meta{options}|}|
\end{quote}

After
\begin{quote}
    |:some attribute = ||{|\meta{options}|}|
\end{quote}
\end{document}
```
image

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-base-external.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-base-external.tex
@@ -305,7 +305,7 @@ is shown below). It is slightly longer, but it can be used in the same way as
     to manually provide the |viewport| option as shown above.
 
     A possible value for |viewport| can be found in the |.pdf| image, search
-    for |/MediaBox = [ ... ]|.
+    for |/MediaBox = ||[ ... ]|.
 \end{key}
 
 
--- a/doc/generic/pgf/text-en/pgfmanual-en-gd-usage-pgf.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-gd-usage-pgf.tex
@@ -522,7 +522,7 @@ of layout keys inside a picture will only open sublayouts.
     layout scopes) and the \meta{end code} at the end of the same \TeX\ scope.
 
     The need for this slightly strange macro arises from the fact that in
-    \tikzname\ we often write things like |[spring layout,node sep=2cm]|. The
+    \tikzname\ we often write things like |[spring layout,node sep=2cm||]|. The
     point is that when the |spring layout| key is executed, we do \emph{not}
     wish to open a layout scope immediately. Rather, this should happen only
     after the option |nodes sep=2cm| has been executed. For this reason,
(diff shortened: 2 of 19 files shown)


hmenke:
Sorry, but I think I have to reject this pull request.

First of all, it seems that you are not using the latest master to compile your example, because for me it renders like this, without missing closing delimiters.

test

The fact that italics are not used for the closing delimiter is something that I am aware of and plan to fix in the future (cf. #831).

You propose to typeset the right-hand side of assignments in upright font, but you are not being consistent (`3cm` is still italics) and you do so by injecting some obscure custom markup into the manual source when this should really be fixed in the parser.  Furthermore, since the beginning the right-hand sides of assignments in the PGF manual have been typeset in italics and I'm not going to change that now.

muzimuzhi:
> First of all, it seems that you are not using the latest master to compile your example, because for me it renders like this, without missing closing delimiters.

Yes you are right. I used the latest master but didn't set pgf repo as a texlive usertree. Following the instructions in https://github.com/pgf-tikz/pgf#development, the missing trailing "]" problem disappears.

So sorry.

------
By the way, not directly related to this pr, there are still two typos I found in doc:

First one, 
https://github.com/pgf-tikz/pgf/blob/357bc05926a057f7c1fcbc95877fc508723ac105/doc/generic/pgf/text-en/pgfmanual-en-library-circuits.tex#L87-L128

This `tikzpicture` is so wide that produces a 200pt+ overfull \hbox
```tex
Overfull \hbox (226.97084pt too wide) in paragraph at lines 87--129
```

Second one, 
https://github.com/pgf-tikz/pgf/blob/357bc05926a057f7c1fcbc95877fc508723ac105/doc/generic/pgf/text-en/pgfmanual-en-tikz-scopes.tex#L361-L380

The example uses option `ultra thick`, but the following explanation mentioned option `thick`.

PR #842, opened 2020-03-23
fixed some manual errors

Mo-Gul:


hmenke:
Undo whitespace changes.

hmenke (changes requested):
Please undo all changes that affect only whitespace.

Mo-Gul:
As always: My editor does the line breaks (and removing trailing spaces) automatically and I want it to do so. So only in `pgfmanual-en-base-animations.tex` it seems I replaced two spaces after an end-of-sentence dot. (I can't remember why I had a look at it.) And while checking that I found (hopefully) the two remaining instances of two spaces in text.

As I have written multiple times (as you), I do these changes for consistency. You are free to undo changes yourself or even reject the pull request. When I am not mistaken we agreed that for **new** files two spaces after an end-of-sentence dot will be used. I will follow this convention (as well).

PR #854, opened 2020-05-14
SVG-driver code for invisibility (nulldevice PS operator)

### LINKED ISSUE #591: dvisvgm backend: issues with beamer standard animations/transitions (\pause)
Hi!
I am trying to translate some PDf slides (LaTeX beamer) to svg to get advantage of the \animate package without needing a non-free PDF viewer. However, I am having some issues with standard beamer animations when using the `dvisvgm` backend, in different scenarios.
This is a simple example that does not work as expected when generating the SVG:

```
\documentclass[dvisvgm,aspectratio=169]{beamer}
\setbeamertemplate{navigation symbols}{}

\begin{document}

\begin{frame}
  \begin{itemize}
  \item 1
    \pause

  \item 2
    \pause

  \item 3
    \begin{block}{foo bar}
      foo bar
    \end{block}
  \end{itemize}
\end{frame}

\end{document}
```

After `latex example.tex; latex example.tex; dvisvgm --font-format=woff --zoom=-1 -p1,- example.dvi` three SVG files are generated, but the beamer block is visible in all of them instead of only in the last one.

I am using **LaTeX-2020** and **dvisvgm 2.9.1**

Regards,
E

agrahn:
As of commit commit https://github.com/mgieseki/dvisvgm/commit/67bf043bf77205557370d2e29e29067938925f9a, `dvisvgm` supports the PostScript `nulldevice` operator. This invalidates / makes redundant the current low-level implementation of the beamer-pause code in the dvisvgm driver of PGF. The proposed pull-request replaces the current implementation and is based on `dvips` driver code which also works with `dvisvgm` now. It also fixes https://github.com/josephwright/beamer/issues/591 .

### DIFF
--- a/tex/generic/pgf/systemlayer/pgfsys-common-svg.def
+++ b/tex/generic/pgf/systemlayer/pgfsys-common-svg.def
@@ -321,10 +321,15 @@
 \def\pgfsys@beveljoin{\pgf@sys@svg@gs{stroke-linejoin="bevel"}}
 
 % Invisibility
-\def\pgfsys@begininvisible{\pgfsysprotocol@literal{<g visibility="hidden">\pgfsys@svg@newline }}
-\def\pgfsys@endinvisible{\pgfsysprotocol@literal{</g>}}
-\def\pgfsys@begininvisiblescope{\pgfsys@beginscope@{ visibility="hidden"}}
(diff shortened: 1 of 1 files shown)


agrahn:
I am closing this PR for now, because there is a side-effect on TikZ graphics on beamer slides with overlays. More testing is required.

Issue #848, opened 2020-04-19
Adding a bit more details to the mindmaps manual


ghost:
There has been repeated confusion about the `concept color` on the Q & A sites. While everything is explained in the manual, I feel it might make sense to stress the distinction between `child` and `node` even more. 

Version: 3.1.5

There is nothing wrong in the manual but the explanations on p. 721 are not too exhaustive. Instead of `Here is an example of a tree built in this way:` we could write

> Let us now build a tree that way. Please note that we pass the `concept color` to the respective `child` and note to a `node` under it.

And in the code example one could add this information, too, since many users just look at the examples.

```latex
% preamble: \usetikzlibrary{mindmap}
\begin{tikzpicture}
\path[mindmap,concept color=black,text=white] node[concept] {Computer Science}
[clockwise from=0]
% note that `sibling angle' can only be defined in % `level 1 concept/.append style={}'
child[concept color=green!50!black] {
node[concept] {practical}
[clockwise from=90]
child { node[concept] {algorithms} }
child { node[concept] {data structures} }
child { node[concept] {pro\-gramming languages} } child { node[concept] {software engineer\-ing} }
}
child[concept color=blue] { % note that the concept color is passed to the child(!)
node[concept] {applied}
[clockwise from=-30]
child { node[concept] {databases} } child { node[concept] {WWW} }
}
child[concept color=red] { node[concept] {technical} } child[concept color=orange] { node[concept] {theoretical} };
\end{tikzpicture}
```

hmenke:
Ha! I'd estimate that 90% of users on Q&A sites don't even know that the manual exists. Improving the manual is always just exercises in futility, but sure it can be done.

hmenke:
@tallmarmot Is the note that @Mo-Gul committed in e6e91c40aa8f929050e9495a08a541c3e0214d4e sufficient?  If yes, please close this and add the closed-fixed label.

ghost:
@hmenke  Looks great except that I had a typo "and note to a" which should be "and not to a", i.e. the "note" should be a "not". Clearly my bad but the typo propagated.

PR #857, opened 2020-05-19
corrected typo in patch of issue #848

Mo-Gul:


hmenke:
Please don't do this on the mainline repo.  Use your own repo to create new branches.

PR #872, opened 2020-06-05
minor manual improvements

Mo-Gul:


doc/generic/pgf/text-en/pgfmanual-en-tikz-shapes.tex
```diff
@@ -1708,11 +1708,11 @@ \subsubsection{Advanced Placement Options}
 \end{key}
 
 
-\subsubsection{Advanced Arrangements of  Nodes}
+\subsubsection{Advanced Arrangements of Nodes}
 
 The simple |above| and |right| options may not always suffice for arranging a
 large number of nodes. For such situations \tikzname\ offers libraries that
-make positioning easier: The |graphdrawing| library and the |matrix| library.
+make positioning easier: The |matrix| library and the |graphdrawing| library.
```
hmenke, line 1715: What's the point of this change?

doc/generic/pgf/text-en/pgfmanual-en-library-matrices.tex
```diff
@@ -12,7 +12,8 @@ \section{Matrix Library}
 
 \begin{tikzlibrary}{matrix}
     This library package defines additional styles and options for creating
-    matrices.
+    matrices. (The basic matrix styles and options can be found in
+    Section~\ref{section-matrices}.)
```
hmenke: This sentence is important enough to stand without parentheses.

doc/generic/pgf/text-en/pgfmanual-en-library-bbox.tex
```diff
@@ -7,158 +7,164 @@
 %
 % See the file doc/generic/pgf/licenses/LICENSE for more details.
 
-\section{Bounding Boxes for B\'ezier Curves}
-
+\section{Bounding Boxes for Bézier Curves}
```
hmenke: You should ask @tallmarmot whether it's okay to make changes to this file.

ghost: I am OK with all but one change but want to mention that I won't be able to edit the file with my editor if it contains letters like `é`.  
The change which I am not OK with concerns the reciprocal. One should explicitly mention the key `/pgf/fpu/install only=reciprocal` in or after line 167. Please note that I ran extensive checks with the `use fpu reciprocal` key but only very little checks with `/pgf/fpu/install only=reciprocal`.

doc/generic/pgf/text-en/pgfmanual-en-tikz-shapes.tex
ghost: In English after a colon one continues lower-case. I'd also avoid doubling the word "library" but write

> make positioning easier: the |matrix| and the |graphdrawing| libraries.

I do not have any preference at concerning the ordering.

ghost (changes requested):
Two requests for changes are added. The one concerning mentioning `/pgf/fpu/install only=reciprocal` explicitly is important.

doc/generic/pgf/text-en/pgfmanual-en-tikz-shapes.tex
Mo-Gul: @hmenke, the point here is the next sentence these libraries are referenced. But in that sentence the order was switched. Now the first mentioned library in the above sentence corresponds to the first reference in the following sentence.

doc/generic/pgf/text-en/pgfmanual-en-library-bbox.tex
Mo-Gul: I had already done that. We agreed that my changes are fine and he will also use this version for further work.

Mo-Gul: Hmm, you didn't mention these changes in the emails we had ...

Regarding `é`: Your editor isn't supporting UTF-8?  
Regarding "reciprocal": If you provide the text, I'll be happy to add it to the file. Is that ok for you?

ghost: @Mo-Gul  I did. I told you explicitly that in the form you propose I cannot agree on publishing this part of the manual. As for the precise wording, this depends on what is written wherever the `/pgf/fpu/install only` is documented. It should be something to the effect that, if there is a `dimension too large` error, the user should add locally `/pgf/fpu/install only=reciprocal` in a scope. You could even add a code example of the type

```
%\usetikzlibrary{bbox}
\begin{tikzpicture}
\pgfmathsetmacro{\rnda}{360*rnd}
\pgfmathsetmacro{\rndb}{360*rnd}
\pgfmathsetmacro{\rndc}{360*rnd}
\pgfmathsetmacro{\rndd}{3*rnd}
\pgfmathsetmacro{\rnde}{3*rnd}
\pgfmathsetmacro{\rndf}{3*rnd}
  \begin{scope}[/pgf/fpu/install only={reciprocal},bezier bounding box=true]
   \draw (0,0)  .. controls (\rnda:\rndd) and (\rndb:\rnde) .. (\rndc:\rndf);
  \end{scope} 
\end{tikzpicture}
```

to make it absolutely clear to the users what they need to do. Notice that in over 99% of the cases one does not seem to need the reciprocal, but I want to make sure that the users can use it also in the other 1%.

hmenke: Using `rnd` in drawings is okay, but then you have to set the random seed before the picture.  Otherwise the compilation might not converge.

ghost: @hmenke You are right. My main "problem" is that, with the updated `fpu` library, I did not find a single problematic case for `bbox`. That is, even though I tested about 10k examples, not a single one had problems. I consider this amazing news!

So how about we add the following code example to the either `fpu` or `decorations` section (I will be happy to edit the source code)

```
%\usetikzlibrary{decorations.markings,fpu}
\begin{tikzpicture}
  \begin{scope}[/pgf/fpu/install only={reciprocal}]
   \draw[postaction=decorate,decoration={markings,
   mark=at position 0.52 with {\draw circle[radius=2pt];}}] 
   plot[smooth,variable=\x,domain=-1:1] (\x*\x*\x,\x*\x);
  \end{scope} 
\end{tikzpicture}
```

and link to this example/section in the `bbox` section? This is one out of a huge set of examples that can be repaired with `fpu`.

hmenke: Having this in the `decorations` section makes the most sense, I believe. @Mo-Gul Could you please add that?

Mo-Gul: Because I didn't know exactly where to place the above code I added it as an extra subsection before section 51.2 (on page 640 in v3.1.5b)

https://github.com/pgf-tikz/pgf/blob/706a9ba576e5cc9cd04bd99f66af3f2b46aa8164/doc/generic/pgf/text-en/pgfmanual-en-library-decorations.tex#L109

because this is of general interest (I think). Any better suggestions?

doc/generic/pgf/text-en/pgfmanual-en-library-matrices.tex
Mo-Gul: done

doc/generic/pgf/text-en/pgfmanual-en-library-bbox.tex
hmenke: What about the UTF-8 issue?  I agree that not using UTF-8 is a bit ridiculous, but if it excludes others from contributing we have to live with ASCII for now.  Not everybody uses fancy editors, like Emacs which implemented UTF-8 support only like 20 years ago.  Some people aren't that modern.

doc/generic/pgf/text-en/pgfmanual-en-library-decorations.tex
```diff
@@ -106,6 +106,28 @@ \subsection{Overview and Common Options}
 \end{key}
 
 
+\subsection{Handling ``Dimension too large error''s}
```
hmenke: That final `s` looks ugly.
```suggestion
\subsection{Handling ``Dimension too large error''}
```

doc/generic/pgf/text-en/pgfmanual-en-library-fpu.tex
```diff
@@ -148,25 +148,30 @@ \subsection{Usage}
...
     You have to be aware of the limitations that the FPU imposes.  It will not
-    magically give TeX better precision, but it will avoid overflow or underflow
-    situation for large or small operands by rescaling them.  In the following
-    example, in the first case the FPU variant performs much better than the
-    normal variant, however, in the second case where a rescaling would not in
-    fact be needed the rescaling introduces a small round-off error.
-    %
-\begin{codeexample}[preamble={\usepgflibrary{fpu}},pre={\pgfkeys{/pgf/fpu=false}}]
+    magically give \TeX{} better precision, but it will avoid overflow or
+    underflow situation for large or small operands by rescaling them.  In the
```
hmenke:
```suggestion
    underflow situations for large or small operands by rescaling them.  In the
```

hmenke (changes requested):
Just some really minor things.  I'm holding this off until I have an approving review by @tallmarmot.

doc/generic/pgf/text-en/pgfmanual-en-library-bbox.tex
Mo-Gul: My suggestion would be: Keep in this file the UTF-8 encoding to be consistent with the rest of the manual.

The *working* file of @tallmarmot is in https://github.com/tallmarmot/pgf/tree/master/experiments/Marmot/bbox. If he agrees he could work with an ASCII version there and if it comes to a pull request again I would again change it to UTF-8. (If necessary I would also happily send him the ASCII version via email which he can commit to his repository as a starting point.) Would that be fine for you, @tallmarmot?

ghost: @Mo-Gul I can open files but cannot save them in the utf-8. (Please note that uploading an utf-8 encoded TeX file to the arXiv is a bit like Russian roulette, the more so since there is really no reason why the source codes should allow these characters. Almost daily there is a question on the sites because someone had some strange character somewhere in their code). So the only problem I see is if I am supposed to edit the manual again, then I'd just change  Bézier back to B\\'ezier so that I can run it, and in the library source code there should not be a problem.

Mo-Gul: (@tallmarmot, I wouldn't generalize that there is no reason for that. For pure code in a programming language this might be true, but here we are talking about "normal text". And there I find it very convenient to write the text with the characters that are needed in the language. Thus I totally prefer to write `"Das Mädchen ist dünn und schön."` instead of `"Das M{\"a}dchen ist d{\"u}nn und sch{\"o}n."` (to give a German example). This is first much more readable and second allows for spell-checking.)

So did I understand you correctly that it is fine for you to keep the UTF-8 encoding here and you work with an ASCII version in your `experiments` folder in your repository?

doc/generic/pgf/text-en/pgfmanual-en-library-decorations.tex
joulev: @hmenke Wouldn't this be better?

```tex
\subsection{Handling ``Dimension too large'' errors}
```

doc/generic/pgf/text-en/pgfmanual-en-library-bbox.tex
ghost: @Mo-Gul As I said, if I ever have to edit this part again, I will have to change it back. This is an English user manual, and not a German poem. (The same logic would prevent you from writing `\int` for an integral sign. As you probably know, unicode math has still plenty of bugs, and frankly I do not see the point of that. I am personally fine with writing `\int` when I need an integral sign, and `\"u` when I need an `û`. And the spell checker does cover this.) But there are good chances that I do not have to edit again. The library code itself should definitely be ASCII, do you agree?

Mo-Gul: Yes, I totally agree that the code should be in ASCII.

hmenke: Could you please get it together?  This is absolutely ridiculous.  We are not going to have a UTF-8/ASCII hybrid that requires manual intervention on every pull, it's either one or the other.  And please stop pushing this arXiv strawman.  This is the PGF manual and not the arXiv.  I really don't care whether some other place doesn't support UTF-8, because we do.

ghost: @hmenke I think we are both wasting our time here. You have changed Bézier to your preferred input, and I am fine with that. All I am saying is that I cannot edit files using this encoding with my editor. This does not mean you need to change it back, and I am not changing my editor, which is very much programmed to suite my needs, and makes me work efficient.

hmenke: I haven't changed anything here.  So ASCII it is.  @Mo-Gul Revert to ASCII.

ghost (dismissed):

doc/generic/pgf/text-en/pgfmanual-en-library-bbox.tex
Mo-Gul: @hmenke, tallmarmot and I are both fine that we keep the UTF-8 version here and he works with the ASCII version in his repository. I have sent him the ASCII version via email.

hmenke: No, that's it, `bbox` is going to be removed from the distribution.  Please find another way to distribute this.  I'm done frying `Extraw{\"u}rste` for you.

ghost:
Do whatever. All I did was to typeset Bézier as B\'ezier. If you want to
join Ulrike's "harass others club and tell them how to edit in
international collaborations", go ahead. Ironically, those who have
absolutely no experience in international collaborations have to have the
strongest voice. And they are happy to throw others under the bus just
because they lack that experience. Fine with me. I am not willing to spend
any more thought on this stuff.

PR #877, opened 2020-06-17
typo?

thinbold:


Mo-Gul (approved):

hmenke (changes requested):
sensible commit message?

hmenke:
No activity. Requested changes are not being addressed.

ilayn:
You can modify the commit message during merge so if the change is valid just push it in.

hmenke:
No, I can't modify the commit message without rebasing.  I can write a custom message for the merge commit but that is not the same.

ilayn:
You can always amend after the merge but yet another two lines and force-push, can be tiring at times.

hmenke:
I went with this now:
```bash
curl -sSL https://github.com/pgf-tikz/pgf/pull/877.patch | sed 's/typo?/Fixed typo: of -> off/g' | git am -s
```
Still it would be much nicer if people just wrote sensible commit messages in the first place.

thinbold:
Sorry about that. Will be careful next time. Have a nice day.

PR #881, opened 2020-06-18
fix for issue #755

Mo-Gul:
(For whatever reason there seem to some commits as well that were already merged, but I don't know why. Either I did something wrong or could the reason be the `rebase` which was done? Do you have an idea? If not, do you have an advice how to "fix" my fork without deleting and reforking it?)

PR #882, opened 2020-06-19
add needed libraries to Lua examples as well (pgf-tikz#640 pgf-tikz#839)

Mo-Gul:
modified Lua (code) `examples` as well as @hmenke suggested.


Mo-Gul:
As I have already written elsewhere I messed up something in my fork when I created the branch (before ilayn suggested another way I should work. So so far only the *new* commits should be
- 8e77219
- 1e13bdf

- [ ] Unfortunately `doc/generic/pgf/version-for-luatex/en/pgfmanual-test.tex` sneaked in the commit as well. Shall I change it back in the next commit or do you take care of that (later)?
- [ ] Further work is to do here is stated elsewhere starting at https://github.com/pgf-tikz/pgf/issues/640#issuecomment-645994493

ilayn:
This looks like all your commits are coming from GitHub website (see the verified marks next to commits). So probably you then rebased and then pushed (I guess). This is rather tricky because if you don't "force push" (another wonderful terminology from git) you get all the previous commits of master as if you committed them in your fork. See all those "merged this that bla bla" commits? Those are precisely the ones which are your attempts to fix the problem as it gets worse and worse :)

I would suggest the following: First type the following

```bash
git branch -v
```

This will list all your branches with the latest commits in them. Since you are struggling with many of them I would say save your work physically (copy and paste changed files to another independent folder somewhere) delete your branches except `master`

```bash
git branch -D <insert_branch_name>  # NOTE: This will delete unconditionally so make sure you saved
```

and start from scratch with my comments from the other issue by first syncing. I think we can afford to create a bit of clutter by closing this and #881 until you get the hang of it. When you are ready you can create a PR cleanly.

Mo-Gul:
(Hopefully I am not stating something wrong now.) Regarding the "Verified": I added the SHA key to TortoiseGit as well following a step-by-step guideline (maybe it was https://dev.to/c33s/sign-your-git-commits-with-tortoise-git-on-windows-3mlf). So locally I am asked at every commit for the GPG password. By this it is *verified* that I really am the cause of the mess ;)

Regarding a fresh start for the `PimpLuaExamples` branch: That is a great idea. I'll 
- rename my clone, 
- (re)clone my fork, 
- delete the `PimpLuaExamples` branch and then
- create it again

before reapplying the changes.

Mo-Gul:
Maybe one more thing: As can be seen the Travis CI failed in the `dvisvgm` path. 

https://travis-ci.com/github/pgf-tikz/pgf/jobs/351457958#L974-L979

I really don't think that I could have messed this up. Can you or @hmenke confirm this? Because for me it is very likely to appear again ...

Issue #839, opened 2020-03-14
[doc] Some examples in Part IV lack codeexample info


muzimuzhi:
Version: 3.1.5b

 * All examples in Sec. 28.3 _Padding and Node Distances_, pp. 422--425. 
Corresponding source file: [`Distances.lua`](https://github.com/pgf-tikz/pgf/blob/master/tex/generic/pgf/graphdrawing/lua/pgf/gd/control/Distances.lua)
 * Sec. 28.4--28.7.3, pp. 426--436, 439--440
 * Sec. 28.8, pp. 442--443
 * Sec. 28.11, pp. 448--449
 * Sec. 30.1.1--30.2, pp. 460--464
 * Sec. 30.3, pp. 466--467
 * Sec. 31.1, pp. 469--470
 * Sec. 31.1.2, pp. 475--476
 * Sec. 31.1.4--31.1.4.5, pp. 477-479
 * Sec. 33--35, pp. 482--488

Related: https://github.com/pgf-tikz/pgf/issues/640#issuecomment-511471922

hmenke:
@Mo-Gul

Mo-Gul:
@hmenke, I would be happy to add the needed stuff, but currently I don't have a clue *how* (see also https://github.com/pgf-tikz/pgf/issues/640#issuecomment-511176605). If you would provide *one* example how to add options to `example` environments in Lua I'll hopefully be able to do the rest.

hmenke:
I'm closing this in favour of #640.  We'll continue to track it over there.

PR #883, opened 2020-06-21
Pimp lua examples (addresses issues #640 #839)

### LINKED ISSUE #640: Examples in the manual as MWE
*Migrated from SourceForge*
Author: krichter722
Timestamp: 2018-07-03 19:40:40.026000

I'm aware that the following is a huge pile of work, however I hope that you devs don't close it immediately because of that. It's intended to make this awesome software easier to use for beginners and trained users which tend to forget or can't keep up with the latest major changes:

It'd be nice to have a link to a [Short, Self Contained, Correct (Compilable), Example](http://sscce.org/) for every code sample so that if you click on a link to a code sample like

```
\begin{tikzpicture}
\draw[step=.5cm,gray,very thin] (-1.4,-1.4) grid(1.4,1.4);
\draw (-1.5,0) -- (1.5,0);
\draw (0,-1.5) -- (0,1.5);
\draw (0,0) circle[radius=1cm];
\draw (3mm,0mm) arc[start angle=0,end angle=30,radius=3mm];
\end{tikzpicture}
```

on page xy you're redirected to the plain text file https://sourceforge.net/p/pgf/wiki/samples/001.tex which contains a SSCCE which can be directly passed to the compiler.

The manual describes what the user needs to do in a satisfying way, starting with

```
\documentclass{article}% say
\usepackage{tikz}
\begin{document}
We are working on
\begin{tikzpicture}
\draw (-1.5,0) -- (1.5,0);
\draw (0,-1.5) -- (0,1.5);
\end{tikzpicture}.
\end{document}
```

However, it's sometimes hard to follow all the packages which have been introduced and settings which have been changed over the course of the chapter where the snippet is shown (which can be dozens of pages long). The plain text links would avoid this and reduce the number of mistakes can make while using the snippet almost to zero.

This would also improve the incapacity of most PDF viewers to copy code accurate, i.e. without introducing unwanted linebreaks and replacing spaces with linebreaks, even though that's not PGF's job.

If you think that's a good idea regardless of the effort, then you might want to keep those SSCCEs coming piece by piece, maybe through merge requests. I can provide some, however, I often find myself trying to get the snippets to compile for hours which is the main reason why I'm suggesting this.

The SSCCEs could be provided for LaTeX, TeX, ConTeX and LuaLaTeX.

### LINKED ISSUE #755: Manual leaves out \usegdlibrary
In the [manual](http://pgf.sourceforge.net/pgf_CVS.pdf), in section 19.2.2 at the bottom of page 265 we can find the line
`For instance, adding tree layout to the above code results in the
following somewhat more pleasing rendering`
. If you have a "\usegdlibrary{trees}" somewhere in your code it works fine, but without it compilation fails with the message:
> ! Package pgfkeys Error: I do not know the key '/tikz/graphs/binary tree layout
> ' and I am going to ignore it. Perhaps you misspelled it. ...

I don't have a Problem with the fact that you need this command, but the documentation does not say that you need this command in this chapter (only 200 pages later) and the error message does not indicate it. It would be nice if you could add a note to the documentation or to the error message

Mo-Gul:
- [ ] Further work to do here is stated elsewhere starting at https://github.com/pgf-tikz/pgf/issues/640#issuecomment-645994493

Closes #640 
Closes #755 
Closes #881


doc/generic/pgf/text-en/pgfmanual-en-tikz-graphs.tex
```diff
@@ -243,8 +243,9 @@ \subsubsection{Concept: Chain Groups}
 
 As can be seen, the placement is not particularly nice by default, use the
 algorithms from the graph drawing libraries to get a better layout. For
-instance, adding |tree layout| to the above code results in the following
-somewhat more pleasing rendering:
+instance, adding |tree layout| to the above code (and
```
ilayn, line 246: So this also closes #881 ?

ilayn (commented):
Don't know much about Lua parts but looks good to me (except that CI error and a minor comment).

hmenke:
The CI failure is a bug in l3backend.

doc/generic/pgf/text-en/pgfmanual-en-tikz-graphs.tex
Mo-Gul: @ilayn, sorry, I thought that I had commented on that already, but it seems that didn't work. So here "again".

You are absolutely right. This is, because I did that change in my forks master branch before following your suggestions to do new changes only in branches. A day after doing these changes I thought that I maybe could have avoided that by checking out the commit before I made that change and creating a branch from there.(?)

I hope that I don't have to do everything again, because this is hopefully a trivial change and therefore doesn't mess up anything when this is applied again.

ilayn: No problem I'll edit this issue text and they will all be closed automagically once this is merged. I think you fought well with git.

ilayn:
I edited the issue description to close the other relevant issues and PRs. Please go over them once if I did the right ones. Then ready when you are ready.

Mo-Gul:
The mentioned issues and pull requests are right, thank you. While the last two could already be merged, the first one will still take a while I guess (as is also mentioned in the first comment here).

hmenke:

> [!TIP]

All tests passed

hmenke (approved):

PR #886, opened 2020-06-26
manual typos and formatting inconsistencies

Mo-Gul:
So far fixed some typos and removed double/multiple spaces.

Things that still can be done to harmonize/correct stuff in manual:
- [x] UK spelling <--> US spelling, e.g. `behavior` vs. `behaviour` --> I prefer US spelling
- [x] `section~` <--> `Section~` --> most of the `\ref` prefixes use the latter, so I'd harmonize to that
- [x] e.g. `$x$ coordinate` <--> `$x$-coordinate`
- [x] `\emph{Note:\/}` <--> `\emph{Note:}`
- [x] e.g. `Mineral\"olprodukte` --> replace by UTF-8 character --> `Mineralölprodukte`

Are there any opinions/arguments against that proposed changes? If not, I would apply them one by one.

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-base-images.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-base-images.tex
@@ -66,7 +66,7 @@ To speedup the compilation, you may wish to use the following class option:
 \begin{command}{\pgfdeclareimage\oarg{options}\marg{image name}\marg{filename}}
     Declares an image, but does not paint anything. To draw the image, use
     |\pgfuseimage{|\meta{image name}|}|. The \meta{filename} may not have an
-    extension. For \textsc{pdf}, the extensions |.pdf|, |.jpg|, and |.png|
+    extension.  For \textsc{pdf}, the extensions |.pdf|, |.jpg|, and |.png|
     will automatically tried. For PostScript, the extensions |.eps|, |.epsi|,
     and |.ps| will be tried.
 
--- a/doc/generic/pgf/text-en/pgfmanual-en-base-images.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-base-images.tex
@@ -66,7 +66,7 @@ To speedup the compilation, you may wish to use the following class option:
 \begin{command}{\pgfdeclareimage\oarg{options}\marg{image name}\marg{filename}}
     Declares an image, but does not paint anything. To draw the image, use
     |\pgfuseimage{|\meta{image name}|}|. The \meta{filename} may not have an
-    extension.  For \textsc{pdf}, the extensions |.pdf|, |.jpg|, and |.png|
+    extension. For \textsc{pdf}, the extensions |.pdf|, |.jpg|, and |.png|
     will automatically tried. For PostScript, the extensions |.eps|, |.epsi|,
     and |.ps| will be tried.
 
--- a/doc/generic/pgf/text-en/pgfmanual-en-library-perspective.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-library-perspective.tex
@@ -190,7 +190,6 @@ projection.
 
     To turn off the perspective in $x$-direction, one must set the $x$ component
     of |p| to \texttt{0} (e.g.\ |p={(0,a,b)}|, where \texttt{a} and \texttt{b}
-
     can be any number and will be ignored). Or one can provide |q| and |r| and
     omit |p|.
 
--- a/doc/generic/pgf/text-en/pgfmanual-en-library-perspective.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-library-perspective.tex
@@ -189,7 +189,7 @@ projection.
     direction is turned off.
 
     To turn of the perspective in $x$-direction, one must set the $x$ component
-    of |p| to \texttt{0} (e.g. |p={(0,a,b)}|, where \texttt{a} and \texttt{b}
+    of |p| to \texttt{0} (e.g.\ |p={(0,a,b)}|, where \texttt{a} and \texttt{b}
     can be any number and will be ignored). Or one can provide |q| and |r| and
     omit |p|.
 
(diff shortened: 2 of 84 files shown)


doc/generic/pgf/text-en/pgfmanual-en-tikz-graphs.tex
```diff
@@ -243,8 +243,9 @@ \subsubsection{Concept: Chain Groups}
 
 As can be seen, the placement is not particularly nice by default, use the
 algorithms from the graph drawing libraries to get a better layout. For
-instance, adding |tree layout| to the above code results in the following
-somewhat more pleasing rendering:
+instance, adding |tree layout| to the above code (and
```
ilayn: didn't you fix this already in #881 ?

Mo-Gul: @ilayn, yes, I did. Have a look at https://github.com/pgf-tikz/pgf/pull/883#discussion_r446149566 for the reason of that.

doc/generic/pgf/text-en/pgfmanual-en-base-images.tex
```diff
@@ -66,7 +66,7 @@ \subsection{Declaring an Image}
 \begin{command}{\pgfdeclareimage\oarg{options}\marg{image name}\marg{filename}}
     Declares an image, but does not paint anything. To draw the image, use
     |\pgfuseimage{|\meta{image name}|}|. The \meta{filename} may not have an
-    extension.  For \textsc{pdf}, the extensions |.pdf|, |.jpg|, and |.png|
+    extension. For \textsc{pdf}, the extensions |.pdf|, |.jpg|, and |.png|
```
hmenke:
```suggestion
    extension.  For \textsc{pdf}, the extensions |.pdf|, |.jpg|, and |.png|
```
This is the correct way to space sentences, see e.g. [Wikipedia](https://en.wikipedia.org/wiki/Sentence_spacing).  Also good text editors like [Emacs](https://www.gnu.org/software/emacs/manual/html_node/emacs/Sentences.html) or [Vim](http://vimdoc.sourceforge.net/htmldoc/options.html#cpo-J) have shortcuts to navigate sentences and therefore need two spaces to distinguish the end of a sentence from the space after “e.g.“ or “i.e.”.

Please revert this and all other similar changes.

hmenke (changes requested):

doc/generic/pgf/text-en/pgfmanual-en-base-images.tex
Mo-Gul: In the stated Wikipedia article there isn't stated that this is the correct way to space sentences. Rather there seems to be opposing opinions and for the newer age it seems to tend to using on space. But that is not the point here because in LaTeX we can decide what we want to see in the output and has nothing to do with the spaces in the input. (I think we agree on that.)

But of course that **editors** use them is a very relevant point, especially when the main contributor to this project uses one of them ;)

I know that you stated that point in the past already and we agreed that I will not touch any of the "two spaces" you have written (for the above reason). And I am pretty sure that I haven't.

So just for clarity: If you insist to revert that changes I will do it. But I just want to mention *my* point again which is *consistency*. In the given example above there are two more sentences before the one that is diffed and there is only given 1 space after the end-of-sentence-dot.

So could you live with the current pull request or shall I change it as you wish? (Please note that this was a one-time-search for the two spaces and thus will not happen again.)

Mo-Gul:
@hmenke, any opinions on the other points I want to change in the manual as stated in the first comment in this pull request?

doc/generic/pgf/text-en/pgfmanual-en-base-images.tex
hmenke: Any news here?

Mo-Gul: I was waiting for your answer which is still lacking. But I'll take it as "you still wish that I revert the end-of-sentence-dots".
[Below](https://github.com/pgf-tikz/pgf/pull/886#issuecomment-650497371) I asked another question which is also not answered yet. Could you please let me know your opinion, too, so I don't do unnecessary/double work. Thank you very much in advance.

hmenke:
4 should definitely be addressed.
1 if you have time and really care about it.
2,3,5 seem very unnecessary.

doc/generic/pgf/text-en/pgfmanual-en-library-perspective.tex
```diff
@@ -188,8 +188,9 @@ \subsection{Defining the perspective}
     Note also that when only |p| is provided, the perspective in $y$ and $z$
     direction is turned off.
 
-    To turn of the perspective in $x$-direction, one must set the $x$ component
+    To turn off the perspective in $x$-direction, one must set the $x$ component
     of |p| to \texttt{0} (e.g.\ |p={(0,a,b)}|, where \texttt{a} and \texttt{b}
+
```
Mo-Gul:
```suggestion
```

doc/generic/pgf/text-en/pgfmanual-en-oo.tex
```diff
@@ -37,7 +37,8 @@ \subsection{Overview}
 Basically, the oo-system supports \emph{classes} (in the object-oriented sense,
 this has nothing to do with \LaTeX-classes), \emph{methods},
 \emph{constructors}, \emph{attributes}, \emph{objects}, \emph{object
-identities}, and (thanks to Saǒ Živanović) \emph{inheritance} and
+identities}, and (thanks to Sašo Živanović) \emph{inheritance} and
+
```
Mo-Gul:
```suggestion
```

hmenke:
Cherry-picked into https://github.com/pgf-tikz/pgf/commit/9ac32f11fd2f8255a8a74b0994cb9aed5c42a3e7

PR #887, opened 2020-06-26
[doc] Fix pdf destinations

### LINKED ISSUE #885: [manual] (some) page link destinations seem to be wrong
I am not 100% sure that this is an TikZ issue. At least in the latest manual (v3.1.5b and current master) some page link destinations seem to be wrong. One example is the link in the first paragraph on page 89

> ... see page 778 for an example).

The number of the destination is correct, but when one clicks on the link one lands "Shading `axis` > `/tikz/top color` > item 4" on page 777 instead of at "Shading `color wheel`" which is indeed on page 778. If it helps, I could add more examples.

The same applies to the PGFPlots manual. Here as an example the page link on page 51

> Please see page 109 ...

jumps to "Section 4.5.12 Scatter Plots" on page 107 instead of to the key `/pgfplots/scatter src` (on indeed page 109).

muzimuzhi:
By stepping a dummy counter in `\pgfmanualentryheadline`, labels used inside environments based on `pgfmanualentry` env now create right pdf destinations.

 - Pros: Only one place is changed.
 - Cons: `\refstepcounter{dummy}` is run 2000+ times, which slows down manual building.
 - Current result: I've manually checked every occurrence of word "page" in the locally built manual, so far so good.

Related: #885

### DIFF
--- a/doc/generic/pgf/macros/pgfmanual-en-macros.tex
+++ b/doc/generic/pgf/macros/pgfmanual-en-macros.tex
@@ -18,7 +18,9 @@
 \definecolor{animationgraphicbackground}{rgb}{0.96,0.96,0.8}
 
 \newenvironment{pgfmanualentry}{\list{}{\leftmargin=2em\itemindent-\leftmargin\def\makelabel##1{\hss##1}}}{\endlist}
-\newcommand\pgfmanualentryheadline[1]{\itemsep=0pt\parskip=0pt{\raggedright\item\strut{#1}\par}\topsep=0pt}
+\newcounter{dummy}
+\newcommand\pgfmanualentryheadline[1]{%
+  \itemsep=0pt\parskip=0pt{\raggedright\item\refstepcounter{dummy}\strut{#1}\par}\topsep=0pt}
 \newcommand\pgfmanualbody{\parskip3pt}
 
 \let\origtexttt=\texttt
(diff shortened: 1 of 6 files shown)


doc/generic/pgf/macros/pgfmanual-en-macros.tex
```diff
@@ -18,7 +18,9 @@
 \definecolor{animationgraphicbackground}{rgb}{0.96,0.96,0.8}
 
 \newenvironment{pgfmanualentry}{\list{}{\leftmargin=2em\itemindent-\leftmargin\def\makelabel##1{\hss##1}}}{\endlist}
-\newcommand\pgfmanualentryheadline[1]{\itemsep=0pt\parskip=0pt{\raggedright\item\strut{#1}\par}\topsep=0pt}
+\newcounter{dummy}
+\newcommand\pgfmanualentryheadline[1]{%
+  \itemsep=0pt\parskip=0pt{\raggedright\item\refstepcounter{dummy}\strut{#1}\par}\topsep=0pt}
```
hmenke: Please give the counter a proper name instead of `dummy`.  How about `\newcounter{pgfmanualentry}`?

hmenke (changes requested):

doc/generic/pgf/macros/pgfmanual-en-macros.tex
muzimuzhi: Sounds good.

hmenke (approved):

PR #892, opened 2020-07-01
Fix for mindmap connection bar glitches

piannucci:
When the distance between two concepts is at least half the sum of the concepts' radii, the connection bar is constructed as expected with initial, bar, and final portions.  There is a visual glitch due to abutting filled path segments on certain anti-aliasing PDF rasterizers, including the one built into macOS, but that could be excused because the apparent gap is truly zero-width and a (sane) printer driver would not produce a white line.  However, there is a more serious visual glitch when the distance between two concepts is less than the sum of the concepts' radii.  In that case, the connection bar path becomes self-intersecting, and the even-odd inclusion test results in unfilled wedges.

This patch unifies the three path segments (initial, bar, and final) into a single continuous filled path segment.  It alters the logic which computes the length of the initial and final flares, proportionally reducing them from their nominal values (of one-half the abutted circle radius) in the case where the distance between concept circles would be too small to fit this shape.

Before:
before

After:
after

As an added bonus, this patch also eliminates the onscreen glitch on anti-aliasing rasterizers.

### DIFF
--- a/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
@@ -17,72 +17,58 @@
 %
 % Parameters: start radius, end radius, amplitude, angle
 
+\newdimen\tikz@lib@mm@startflarelength
+\newdimen\tikz@lib@mm@endflarelength
 \pgfdeclaredecoration{circle connection bar}{initial}
 {%
-  \state{initial}[width=0pt,next state=bar]
+  \state{initial}[width=0pt,next state=final]
   {
+    \pgfmathsetlength\pgfutil@tempdimb{\pgfdecorationsegmentamplitude}%
     {
     \pgftransformxshift{-\pgfkeysvalueof{/pgf/decoration/start radius}}%
-    \pgfpathmoveto{\pgfpointpolar{\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/start radius}}}
-    \pgfpatharc{\pgfdecorationsegmentangle}{-\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/start radius}}
-    \pgfutil@tempcnta=-\pgfdecorationsegmentangle\relax
-    \advance\pgfutil@tempcnta by90\relax
     \pgfmathsetlength\pgfutil@tempdima{\pgfkeysvalueof{/pgf/decoration/start radius}}
-    \pgfmathsetlength\pgfutil@tempdimb{\pgfdecorationsegmentamplitude}
-    \pgfpathcurveto
-    {\pgfpointadd
-      {\pgfpointpolar{-\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/start radius}}}
-      {\pgfpointpolar{\the\pgfutil@tempcnta}{.25\pgfutil@tempdima}}}
-    {\pgfqpoint{1.25\pgfutil@tempdima}{-.5\pgfutil@tempdimb}}
-    {\pgfqpoint{1.5\pgfutil@tempdima}{-.5\pgfutil@tempdimb}}
-    \pgfpathlineto{\pgfpoint{1.5\pgfutil@tempdima}{.5\pgfutil@tempdimb}}
+    \pgfpathmoveto{\pgfpoint{\pgfutil@tempdima+1.0\tikz@lib@mm@startflarelength}{.5\pgfutil@tempdimb}}
     \pgfutil@tempcnta=\pgfdecorationsegmentangle\relax
     \advance\pgfutil@tempcnta by-90\relax
     \pgfpathcurveto
-    {\pgfpoint{1.25\pgfutil@tempdima}{.5\pgfutil@tempdimb}}
+    {\pgfpoint{\pgfutil@tempdima+0.5\tikz@lib@mm@startflarelength}{.5\pgfutil@tempdimb}}
     {\pgfpointadd
       {\pgfpointpolar{\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/start radius}}}
       {\pgfpointpolar{\the\pgfutil@tempcnta}{.25\pgfutil@tempdima}}}
     {\pgfpointpolar{\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/start radius}}}
-    \pgfpathclose
-    }
-  }%
-  \state{bar}[width=0pt,next state=end]
-  {
-    \pgfmathsetlength\pgfutil@tempdima{\pgfkeysvalueof{/pgf/decoration/start radius}}%
-    \pgfmathsetlength\pgfutil@tempdimb{\pgfkeysvalueof{/pgf/decoration/end radius}}%
-    \pgfmathsetlength\pgf@xc{\pgfdecorationsegmentamplitude}%
-    \pgfpathrectangle
-    {\pgfqpoint{.5\pgfutil@tempdima}{-.5\pgf@xc}}
-    {\pgfpoint{\pgfdecoratedremainingdistance+-.5\pgfutil@tempdimb+-.5\pgfutil@tempdima}{\pgf@xc}}
-  }%
-  \state{end}[width=0pt,next state=final]
-  {
-    {
-    \pgftransformxshift{\pgfdecoratedremainingdistance}%
-    \pgftransformxscale{-1}%
-    \pgftransformxshift{-\pgfkeysvalueof{/pgf/decoration/end radius}}%
-    \pgfpathmoveto{\pgfpointpolar{\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/end radius}}}
-    \pgfpatharc{\pgfdecorationsegmentangle}{-\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/end radius}}
+    \pgfpatharc{\pgfdecorationsegmentangle}{-\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/start radius}}
     \pgfutil@tempcnta=-\pgfdecorationsegmentangle\relax
     \advance\pgfutil@tempcnta by90\relax
-    \pgfmathsetlength\pgfutil@tempdima{\pgfkeysvalueof{/pgf/decoration/end radius}}
-    \pgfmathsetlength\pgfutil@tempdimb{\pgfdecorationsegmentamplitude}%
     \pgfpathcurveto
     {\pgfpointadd
-      {\pgfpointpolar{-\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/end radius}}}
+      {\pgfpointpolar{-\pgfdecorationsegmentangle}{\pgfkeysvalueof{/pgf/decoration/start radius}}}
       {\pgfpointpolar{\the\pgfutil@tempcnta}{.25\pgfutil@tempdima}}}
-    {\pgfqpoint{1.25\pgfutil@tempdima}{-.5\pgfutil@tempdimb}}
-    {\pgfqpoint{1.5\pgfutil@tempdima}{-.5\pgfutil@tempdimb}}
-    \pgfpathlineto{\pgfpoint{1.5\pgfutil@tempdima}{.5\pgfutil@tempdimb}}
+    {\pgfpoint{\pgfutil@tempdima+0.5\tikz@lib@mm@startflarelength}{-.5\pgfutil@tempdimb}}
+    {\pgfpoint{\pgfutil@tempdima+1.0\tikz@lib@mm@startflarelength}{-.5\pgfutil@tempdimb}}
+    }
+    {
+    \pgftransformxshift{\pgfdecoratedremainingdistance}%
+    \pgftransformscale{-1}%
+    \pgftransformxshift{-\pgfkeysvalueof{/pgf/decoration/end radius}}%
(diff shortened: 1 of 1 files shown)


tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
```diff
@@ -110,17 +101,41 @@
 }}%
 \tikzset{every circle connection bar/.style={}}%
 
+\pgfkeys{/pgf/decoration/start flare length/.initial=0pt}%
+\pgfkeys{/pgf/decoration/end flare length/.initial=0pt}%
```
hmenke: When you introduce new parameters you have to write documentation.

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
```diff
@@ -110,17 +101,41 @@
...
   \pgf@xa=\pgf@x%
   \pgf@process{\pgfpointtransformed{\pgfpointanchor{\tikztotarget}{west}}}%
   \advance\pgf@xa by-\pgf@x%
   \pgfkeys{/pgf/decoration/end radius/.expanded=\the\pgf@xa}%
+  \advance\pgfutil@tempdima by.5\pgf@xa\relax%
+  % Now tempdima = .5 start radius + .5 end radius = nominal sum of flare lengths.
+  \pgfpointdiff{\pgfpointanchor{\tikztostart}{center}}%
+               {\pgfpointanchor{\tikztotarget}{center}}%
+  \pgfmathparse{veclen(\the\pgf@x,\the\pgf@y)}%
+  \pgfmathsetlength\pgfutil@tempdimb{\pgfmathresult}%
```
hmenke: Don't use these macros in low level code.  They are slow and pollute the log file when tracing.
```suggestion
  \pgfmathveclen@{\pgf@sys@tonumber\pgf@x}{\pgf@sys@tonumber\pgf@y}%
  \pgfutil@tempdimb=\pgfmathresult pt
```

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
```diff
@@ -110,17 +101,41 @@
...
   \pgf@process{\pgfpointtransformed{\pgfpointanchor{\tikztostart}{west}}}%
   \advance\pgf@xa by-\pgf@x%
   \pgfkeys{/pgf/decoration/start radius/.expanded=\the\pgf@xa}%
+  \pgfutil@tempdima=.5\pgf@xa\relax%
   \pgf@process{\pgfpointtransformed{\pgfpointanchor{\tikztotarget}{center}}}%
   \pgf@xa=\pgf@x%
   \pgf@process{\pgfpointtransformed{\pgfpointanchor{\tikztotarget}{west}}}%
   \advance\pgf@xa by-\pgf@x%
   \pgfkeys{/pgf/decoration/end radius/.expanded=\the\pgf@xa}%
+  \advance\pgfutil@tempdima by.5\pgf@xa\relax%
```
hmenke: This is not safe. The `\pgfutil@tempdim` registers should be used immediately after assignment because any other macro may use them as well.

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
```diff
@@ -110,17 +101,41 @@
...
+  \pgfmathparse{veclen(\the\pgf@x,\the\pgf@y)}%
+  \pgfmathsetlength\pgfutil@tempdimb{\pgfmathresult}%
+  \advance\pgfutil@tempdimb by-2\pgfutil@tempdima%
+  % Now tempdimb = room available for flares.
+  \ifdim\pgfutil@tempdimb>\pgfutil@tempdima\relax%
+    \pgfutil@tempdimb=\pgfutil@tempdima\relax%
+  \fi%
+  % Now tempdimb = actual sum of flare lengths.
+  \pgf@xa=\pgfkeysvalueof{/pgf/decoration/start radius}\relax%
+  \pgf@xa=\dimexpr .5\pgf@xa * \pgfutil@tempdimb / \pgfutil@tempdima\relax%
```
hmenke: This does not do what you expect.  Try for example this:
```latex
\pgfutil@tempdima=1pt
\pgfutil@tempdimb=2pt
\the\dimexpr \pgfutil@tempdima / \pgfutil@tempdimb \relax % 0.00002pt
```

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
```diff
@@ -17,72 +17,63 @@
...
-    \pgfpathrectangle
-    {\pgfqpoint{.5\pgfutil@tempdima}{-.5\pgf@xc}}
-    {\pgfpoint{\pgfdecoratedremainingdistance+-.5\pgfutil@tempdimb+-.5\pgfutil@tempdima}{\pgf@xc}}
-  }%
-  \state{end}[width=0pt,next state=final]
-  {
     {
     \pgftransformxshift{\pgfdecoratedremainingdistance}%
     \pgftransformxscale{-1}%
+    \pgftransformyscale{-1}%
```
hmenke: This will reset the xscale.  Did you mean this?
```suggestion
    \pgftransformscale{-1}%
```

hmenke (changes requested):
I haven't reviewed this in too much detail yet, but at first sight this definitely needs some more work.

ilayn:
Also how does this behave for gradient linkages such as the ones given just before section 60.4 in the manual?

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
piannucci: OK, I'm re-writing to use globals similarly to the the way `\tikz@parse@switch` does.  Please excuse my ignorance: is there any reason why I should not use a pair of `\newdimen`?  I ask because I see a lot of re-use of the same temporary variables, which makes me wonder if maybe variable allocations are a significant performance problem in this environment.

piannucci: To be clear, I do not intend for these to be customizable; I was just following the pattern of the code that marshals `/pgf/decoration/start radius` and `end radius` into the decoration macro.

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
piannucci: Thank you.  I was in a hurry and I didn't have enough familiarity with this environment.

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
piannucci: OK, but this makes me doubly interested in the answer to my question above about whether newdimen is problematic, since it seems to be used so infrequently.

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
hmenke: The avoidance of `\newdimen` dates back to the pre-eTeX times when there used to be only 256 dimen registers, so you wouldn't want to waste more than necessary.  In eTeX there are 32767 dimen registers, so running out is no longer a concern.

hmenke: Even if those parameters are not supposed to be modified, they are still part of the user interface and therefore require documentation.  See here:
https://github.com/pgf-tikz/pgf/blob/a9818b15a158f15a15cffcf7907c26f3f487005e/doc/generic/pgf/text-en/pgfmanual-en-library-decorations.tex#L80-L96

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
hmenke: This is probably safe.
```suggestion
  \pgf@xa=.5\pgf@xa
  \pgfmathdivide@{\pgf@sys@tonumber\pgfutil@tempdimb}{\pgf@sys@tonumber\pgfutil@tempdima}%
  \pgf@xa=\pgfmathresult\pgf@xa
```

hmenke: Oh wait, your version actually *does* work because of a corner case in `\dimexpr` where it evaluates `a * b / c` without returning the intermediate result to TeX.  But if you swap it around, i.e.
```latex
\dimexpr .5\pgf@xa / \pgfutil@tempdima * \pgfutil@tempdimb\relax
```
the result is wrong.  Maybe we can leave your version in this instance, but generally I'm not a huge fan of `\dimexpr` (and `\numexpr` for that matter).

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
piannucci: OK, I found a better way.

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
piannucci: I've rewritten it using pgfmathdivide@, multiply, and add.

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
piannucci: One line is better than two, so, good.  But FWIW my experiments don't match your expectation.  If the x scale were reset, then I would expect commenting out line 52 would do nothing; but that is not the case.

piannucci:
I've rewritten the patch so that it does not introduce any new parameters.  I've also tried to do a better job of understanding TeX's scoping rules; I think this patch should be more robust.

Some things are still baffling to me: why does `\pgfmathdivide@` consistently return 1.0 if the calls to `\pgf@sys@tonumber` are inline in the parameters, rather than edef'd first?

piannucci:
@ilayn It seems to work fine.

hmenke:
> Some things are still baffling to me: why does \pgfmathdivide@ consistently return 1.0 if the calls to \pgf@sys@tonumber are inline in the parameters, rather than edef'd first?

Can't happen. What are your input parameters?

hmenke:
pgfmath uses the same registers internally.

https://github.com/pgf-tikz/pgf/blob/a9818b15a158f15a15cffcf7907c26f3f487005e/tex/generic/pgf/math/pgfmathutil.code.tex#L96-L104

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymindmap.code.tex
```diff
@@ -121,6 +107,33 @@
...
+    \pgfmathsetlength\pgf@y{\pgfkeysvalueof{/pgf/decoration/end radius}}%
+  }%
+  \advance\pgf@x by\pgf@y\relax% x = sum of radii = 2 * nominal sum of flare lengths
+  \advance\pgf@xa by-\pgf@x\relax% xa = distance between circles = room available for flares
+  \pgf@x=.5\pgf@x\relax% x = nominal sum of flare lengths
+  \ifdim\pgf@xa>\pgf@x\relax%
+    \pgf@xa=\pgf@x\relax%
+  \fi% xa = actual sum of flare lengths
+  \edef\num{\pgf@sys@tonumber\pgf@xa}%
+  \edef\denom{\pgf@sys@tonumber\pgf@x}%
```
hmenke: This is the worst antipattern ever.  Just clobbering user-accessible names without even checking whether they are used.

piannucci: I thought the scoping rules prevented those definitions from leaking out into user scope.  If I had enclosed lines 125-127 in curly braces, would I have been correct?

piannucci: Or do I have to worry about the possibility that those names have already been allocated to registers of some special type?  Please understand that TeX is not one of my primary languages.  I seem to be encountering a lot of "gotchas".

hmenke: Which scope?  There is no enclosing scope around it which means that the scope is the entire TikZ path and someone can very well use user macros on a frontend path.

piannucci:
Temporary variable names changed to something unique.

hmenke:

> [!TIP]

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-mindmaps`
          - `pgfmanual-en-tutorial-map`

        </details>

    </details>

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-mindmaps`
          - `pgfmanual-en-tutorial-map`

        </details>

    </details>

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-mindmaps`
          - `pgfmanual-en-tutorial-map`

        </details>

    </details>

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-mindmaps`
          - `pgfmanual-en-tutorial-map`

        </details>

    </details>

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-mindmaps`
          - `pgfmanual-en-tutorial-map`

        </details>

    </details>

</details>

PR #899, opened 2020-07-18
3d text along path

ZhiyuanLck:
Declare a new decoration: `3d text along path`

Use this decoration to put text on curved surface.

Only two keys are added:
- `3d raise`: apply correct y shift
- `yvec`: specify the direction of y unit vector by an angle or end point(start point is the origin point of the segment)

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-library-decorations.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-library-decorations.tex
@@ -1570,6 +1570,66 @@ shapes. This library is included mostly for historical reasons, using the
     \end{key}
 \end{decoration}
 
+\begin{decoration}{3d text along path}
+    This decoration is almost the same to the |text along path| decoration
+    except that each character is placed on different coordinate plane
+    specified by two unit vectors. The $x$-direction is the tangent direction
+    of the point where the character locates on the path. The $y$-direction is
+    a fixed direction or the direction from the point where the character
+    locates to a fixed point.
+    %
+\begin{codeexample}[preamble={\usetikzlibrary{decorations.text}}]
+\begin{tikzpicture}[decoration={3d text along path,
+  text={|\large\color{red}|3d|| text along path!},
+  text align=center, 3d raise=1cm-.5ex, yvec=60}]
+
+  \draw [fill=gray!30]
+    (0, 2) arc (120:0:2cm) -- ++(60:2cm) arc (0:120:2cm) -- cycle;
+  \path [postaction={decorate}] (0,2) arc (120:0:2cm);
+\end{tikzpicture}
+\end{codeexample}
+\end{decoration}
+
+    \begin{key}{/pgf/decoration/yvec=\marg{number or point} (initially 90)}
+        Set the $y$-direction of every character coordinate plane. If a number
+        is provided, the $y$-direction is the \meta{number}. If a point is
+        provided, the $y$-direction is the direction from the point where the
+        character locates to this \meta{point}.
+        %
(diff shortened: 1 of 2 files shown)


muzimuzhi:
The first two commits are avoidable. You can either rebase these three commits into single one, or cherry-pick the third one, then force push. And, how about adding documentation for newly added options as well?

ZhiyuanLck:
I have tried to add the documentation.

2020-07-26 13-09-03 的屏幕截图

2020-07-26 13-11-49 的屏幕截图

doc/generic/pgf/text-en/pgfmanual-en-library-decorations.tex
```diff
@@ -1570,6 +1570,66 @@ \subsection{Text Decorations}
     \end{key}
 \end{decoration}
 
+\begin{decoration}{3d text along path}
+    This decoration is almost the same to the |text along path| decoration
+    except that each character is placed on different coordinate plane
```
Mo-Gul, line 1575:
```suggestion
    This decoration is almost the same as the |text along path| decoration
    except that each character is placed on a different coordinate plane
```

doc/generic/pgf/text-en/pgfmanual-en-library-decorations.tex
```diff
@@ -1570,6 +1570,66 @@ \subsection{Text Decorations}
...
+  \draw [fill=gray!30]
+    (0, 2) arc (120:0:2cm) -- ([shift={(60:4cm)}]1.5,1) -- cycle;
+  \path [postaction={decorate}] (0,2) arc (120:0:2cm);
+\end{tikzpicture}
+\end{codeexample}
+    \end{key}
+
+    \begin{key}{/pgf/decoration/3d raise=\marg{number or dimension} (initially
+      0pt)}
+        The characters are raised in the $y$-direction of every charater
```
Mo-Gul, line 1613:
```suggestion
        The characters are raised in the $y$-direction of every character
```

ZhiyuanLck: Thanks for review！But I am afraid that I can't push a new commit until 10 days later... I'll do it as soon as I go home.

hmenke (changes requested):
First of all thank you for your contribution.  Unfortunately, I see a couple of issues with this code:

1. This does not belong into `tikzlibrarydecorations.text.code.tex` but should go into `pgflibrarydecorations.text.code.tex`, so that raw PGF users can benefit as well.

2. Almost the entire decoration is a straight copy-paste from the existing `text along path` decoration.  Do you think this massive code-duplication is a good idea, in particular in the context of long-term maintainability?

3. At least the newly introduced `3d raise` parameter seem superfluous to me and could (should?) be replaced by the exisitng `raise` option.  The `yvec` parameter is also not very descriptive of what it does.

ZhiyuanLck:
Thanks for your advise! They are very helpful for me. And I am sorry for my obstrusion. Should I revise the code of `text along path` and add `3d` option to control the appearance of the text?

 I have thought about integrating `3d raise` in `raise`, but `raise` can be applied to all decorations but `3d raise` seems to be meaningful only here, so I give up.

ZhiyuanLck:
If there is no need to keep original `raise` with `3d text along path`, I am happily to replace `3d raise` with `raise`.

hmenke:
As far as I can see the `raise` option is currently unused for `text along path`, so you can just hijack it.

hmenke:
Are you still working on this?

ZhiyuanLck:
No, sorry to foget it. I'll close it and reopen it only if someone really need this fearture.

PR #920, opened 2020-09-07
Fix math lib

### LINKED ISSUE #847: math library can't cope  with predefined, non-robust variables
**Brief outline of the bug**

When a variable has a predefined, non robust definition using it in the math library breaks.

```latex
\documentclass[]{article}
\usepackage{tikz}
\usetikzlibrary{math}

\newcommand\x{blub} %breaks
%\DeclareRobustCommand\x{blub} % works
\begin{document}
        \begin{tikzpicture}[
            evaluate={function f(\x) {return 1;};},
            ]
        \node at (1,1){x}; \node at (1,{f(1)}) {n}; %
        \end{tikzpicture}
\end{document}
```
this gives the error

~~~~
! Package tikz Error: tikz math library: Unknown function or keyword 'blub=1;re
turn'.
~~~~

A robust definition seems to work fine, but this naturally not something one can rely on. 

The problem is connected to problems with the metropolis theme, which uses internally with the unicode engines `\font\x= ...` to check for fonts, which then makes it impossible to use `\x` in math expressions.

muzimuzhi:

In TikZ's `math` lib, if `\x` is
 - a macro with prefix(ex) in its `\meaning` output, for example defined by `\long\def\x{}` or
 - a font declaration e.g., defined by `\font\x=cmr10`,

then using `\x` as variable name raises errors. This PR fixes the problem. As mentioned in https://github.com/pgf-tikz/pgf/issues/847#issuecomment-688018512, there are more corner cases like font declarations that currently `math` lib won't handle.

Test case:
```tex
\documentclass{article}
\usepackage{tikz}
\usetikzlibrary{math}

\begin{document}
\long\def\x{blub}
\font\y=cmr10 at 10pt\selectfont

% expected output: "1, 2, 3"
\tikzmath{
  \x = 1;
  \y = 2;
  function f(\x, \y) {return \x + \y;};
  int \z;
  \z = f(\x,\y);
  print {\x, \y, \z};
}
\end{document}
```

Fixes #847 

**Checklist**

### DIFF
--- a/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymath.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarymath.code.tex
@@ -25,9 +25,21 @@
     \if#1u% undefined
         \let\tikz@math@meaning=\tikz@math@meaning@macro%
     \else%
-        \if#1m%
+        \if#1m% "macro:->[...]"
             \let\tikz@math@meaning=\tikz@math@meaning@macro%
         \else%
+            \if#1s% "select font [...]", treat like a macro
+                \let\tikz@math@meaning=\tikz@math@meaning@macro
+            \else
+                \if#2l% "\long [...]macro:->[...]"
+                    \let\tikz@math@meaning=\tikz@math@meaning@macro
+                \else
+                    \if#2o% "\outer [...]macro:->[...]"
+                        \let\tikz@math@meaning=\tikz@math@meaning@macro
+                    \else
+                        \if#2p% "\protected [...]macro:->[...]"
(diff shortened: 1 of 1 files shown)


hmenke:
I appreciate your enthusiasm.  Unfortunately, this patch also has problems.  It fails if the escape character is not the backslash and also on active characters.  It's possible to make this work, but I won't have time for about another two weeks, so I will have to hold this off, sorry.

hmenke:
If you want to get a sneak-peak of how I'm going to do it, look at the `\token_if_` family of macros in `expl3`.  I will also put this in PGF instead of TikZ because similar functionality is used in pgfmath.

PR #937, opened 2020-10-10
Consistently define `.@cmd` long and fix spurious spaces

muzimuzhi:

This PR fixes two minor problems I found when reading `pgfkeys.code.tex`.
 1. In definition of `/.add code`, branch that subkey `/.@args` is defined, the subkey `/.@cmd` is redefined short (without prefix `\long`), which is inconsistent with all other codes that (re)define `/.@cmd`.
 1. Some spurious spaces found in defining five `/errors/...` keys. 

PS: With `\unexpanded` from eTeX, the code for `/errors/...` keys can be simplified so that the use of `\toks` is avoidable.

PS2: To avoid "at" github user cmd, `` `/.@cmd` `` is used instead of `"/.@cmd"` in commit message.

**Checklist**

### DIFF
--- a/tex/generic/pgf/utilities/pgfkeys.code.tex
+++ b/tex/generic/pgf/utilities/pgfkeys.code.tex
@@ -1035,50 +1035,50 @@
 
 \pgfkeys{
   /errors/boolean expected/.code 2 args=%
-  {
-    \toks1={#1}
-    \toks2={#2}
+  {%
+    \toks1={#1}%
+    \toks2={#2}%
     \pgfkeys@error{%
       Boolean parameter of key '\the\toks1' must be 'true' or 'false', not
       '\the\toks2'. I am going to ignore it%
-    }
+    }%
   },
   /errors/value required/.code 2 args=%
-  {
-    \toks1={#1}
+  {%
+    \toks1={#1}%
     \pgfkeys@error{%
       The key '\the\toks1' requires a value. I am going to ignore this
       key%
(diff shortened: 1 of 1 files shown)


tex/generic/pgf/utilities/pgfkeys.code.tex
```diff
@@ -749,7 +749,7 @@
   {%
     \pgfkeysgetvalue{\pgfkeyscurrentpath/.@args}{\pgfkeys@tempargs}%
     \pgfkeysgetvalue{\pgfkeyscurrentpath/.@body}{\pgfkeys@tempbody}%
-    \def\pgfkeys@marshal{\expandafter\gdef\expandafter\pgfkeys@global@temp\pgfkeys@tempargs}%
+    \def\pgfkeys@marshal{\expandafter\long\expandafter\gdef\expandafter\pgfkeys@global@temp\pgfkeys@tempargs}%
```
hmenke, line 752: You don't actually need an `\expandafter` before `\long`.  This would just work fine as well.
```suggestion
    \def\pgfkeys@marshal{\long\expandafter\gdef\expandafter\pgfkeys@global@temp\pgfkeys@tempargs}%
```

muzimuzhi: Ah yes. I might be affected/misled by other two `\expandafter\long` in `pgfkeys.code.tex`.

PR #961, opened 2020-12-17
Document \pgfpointtransformed and the relation of /.code and /.initial

### LINKED ISSUE #654: /.code keys don't respect /.initial
If a keys has been defined with `/.code`, the initial value can't be set with `/.initial`.  The following MWE shoud throw an undefined control sequence error when trying to set `/.initial` but that does not happen.
```latex
\input pgfkeys.tex
\pgfkeys{foo/.code=\ERROR}
\pgfkeys{foo/.initial=bar}
%\pgfkeys{foo=bar} % BOOM
\bye
```
This behaviour should either be fixed or documented.

### LINKED ISSUE #844: `\pgfpointtransformed` is mentioned, but not documented
Version: 3.1.5b

On page 1164, the documentation to `\pgfpointtransformednonlinear` reads:

> Works like `\pgfpointtransformed`, but also applies the current nonlinear transformation...

However, `\pgfpointtransformed` is not documented at all - but it really should be, I'd say it is rather a rather important function.

For developers, it would also be nice to know what the recommended way for generic code (i.e. without knowing what the user specifies as a transformation) is: Always call  the (more general?) nonlinear point transformation, or, as `\pgfpathmoveto` or `\pgfpathlineto`, always only call the linear one and execute special macros afterwards that deal with the nonlinear case (which probably then require some patching of `\pgftransformnonlinear` in order to install these commands.

muzimuzhi:

Two documentary changes:
 - added doc for `\pgfpointtransformed` which is adapted from its [code comments](https://github.com/pgf-tikz/pgf/blob/c3fb4153410ebc1a0491df7c0189050d43c8ba7a/tex/generic/pgf/basiclayer/pgfcorepoints.code.tex#L90-L100)
 - clarify that a `pgfkeys` key can be defined to both store a value and execute commands, and if so `<key>=<value>` always execute commands.

To reduce the number of merging commits, I combined them into a single PR.

Fixes #844
Fixes #654

**Checklist**

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-pgfkeys.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-pgfkeys.tex
@@ -1253,6 +1253,22 @@ directly stored in a key.
     and, thus, |\pgfkeysnovalue| will be stored in |/my key|.
 
     To retrieve the value stored in a key, the handler |/.get| is used.
+    
+    \medskip
+    \emph{Remark:} A key can both store a value and execute commands. If so,
+    using \meta{key} will \emph{always} execute commands with the passed
+    value, or the default value |\pgfkeysnovalue| if no value is provided.
+    Note that the stored value is never used. To update the stored value, the
+    handler |/.initial| or command |\pgfkeyssetvalue| is used.
+    %
+\begin{codeexample}[]
+\pgfkeys{/my key/.initial=red}
+\pgfkeys{/my key/.code=#1}
+% "/my key" now both stores the value "red" and executes commands
+\pgfkeys{/my key=blue}
+% "/my key" now still stores the value "red"
+\end{codeexample}
+    %
 \end{handler}
 
 \begin{handler}{{.get}|=|\meta{macro}}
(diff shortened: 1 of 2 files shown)


doc/generic/pgf/text-en/pgfmanual-en-pgfkeys.tex
```diff
@@ -1253,6 +1253,22 @@ \subsubsection{Defining Value-, Macro-, If- and Choice-Keys}
     and, thus, |\pgfkeysnovalue| will be stored in |/my key|.
 
     To retrieve the value stored in a key, the handler |/.get| is used.
+    
+    \medskip
+    \emph{Remark:} A key can both store a value and execute commands. If so,
+    using \meta{key} will \emph{always} execute commands with the passed
+    value, or the default value |\pgfkeysnovalue| if no value is provided.
+    Note that the stored value is never used. To update the stored value, the
+    handler |/.initial| or command |\pgfkeyssetvalue| is used.
```
hmenke, line 1265: Please add that this will not be changed for compatibility.  Because you did fix it at some point but that broke other uses and we had to roll back.

hmenke (changes requested):

doc/generic/pgf/text-en/pgfmanual-en-pgfkeys.tex
muzimuzhi: Added a footnote.

hmenke (approved):

PR #967, opened 2020-12-21
Handle \relax and frozen \relax in tikz path

### LINKED ISSUE #966: Suggestion to comment in the manual on using `\ifnum` in paths
This is not a real issue, but a suggestion. It may have to do with a LaTeX kernel change, but I am not really sure (yet I would have thought that this problem wasn't there with some older kernel). Anyway, consider the example
```
\documentclass[border=3mm,tikz]{standalone} 
\begin{document}
\begin{tikzpicture}
 \path foreach \X in {1,...,6}
 {(\X,0) node[circle,draw](\X){}}
 foreach  \X [evaluate=\X as \Xm using {int(\X-1)}]  in {1,...,6}
 {foreach  \Y in {1,...,6}
 {\ifnum\Y<\X
 \ifnum\Y=\Xm 
   (\Y) edge[->] (\X) 
 \else  
  \ifodd\Y
  (\Y) edge[->,bend left] (\X) 
  \fi
 \fi
 \fi}};
\end{tikzpicture}
\end{document}
```
On my updated TeXLive2020 installation this yields an error. (I totally understand that this is a stupid way to get the output, it is just a small example.) According to what I find, these errors can be avoided by adding seemingly useless `\numexpr` "commands". In this example, replacing `\ifnum\Y<\X` by `\ifnum\Y<\numexpr\X` does the trick. 
```
\documentclass[border=3mm,tikz]{standalone} 
\begin{document}
\begin{tikzpicture}
 \path foreach \X in {1,...,6}
 {(\X,0) node[circle,draw](\X){}}
 foreach  \X [evaluate=\X as \Xm using {int(\X-1)}]  in {1,...,6}
 {foreach  \Y in {1,...,6}
 {\ifnum\Y<\numexpr\X
 \ifnum\Y=\Xm 
   (\Y) edge[->] (\X) 
 \else  
  \ifodd\Y
  (\Y) edge[->,bend left] (\X) 
  \fi
 \fi
 \fi}};
\end{tikzpicture}
\end{document}
```
So this "issue" is to propose to add a comment in the manual that informs users about the possibility to potentially fix nested (?) `\ifnum`s by adding `\numexpr` in places where they do not seem to be required.

muzimuzhi:

As suggested in https://github.com/pgf-tikz/pgf/issues/966#issuecomment-749154057, this PR makes `\relax` and the frozen `\relax` handled in tikz path parsing. An intro to frozen `\relax` can be found in [this TeX-SX answer](https://tex.stackexchange.com/a/57417) and [`texdoc interface3`, sec. XVI.7](https://github.com/latex3/latex3/blob/c297e780ff706dab7b30f9ad5153f2f58f542de9/l3kernel/l3token.dtx#L1162-L1164).

PS: Perhaps `\tikz@frozen@relax@token` could be put in `pgfutil-common.tex`.

Fixes #966 (or, implements)

**Checklist**

### DIFF
--- a/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
@@ -2587,7 +2587,7 @@
 
 \def\tikz@skip#1{\tikz@scan@next@command#1}%
 \def\tikz@expand{%
-  \advance\tikz@expandcount by -1%
+  \advance\tikz@expandcount by -1
   \ifnum\tikz@expandcount<0\relax%
     \tikzerror{Giving up on this path. Did you forget a semicolon?}%
     \let\pgfutil@next=\tikz@finish%
@@ -2960,7 +2960,7 @@
 % -| <point>
 
 \def\tikz@hv@lineto{%
-  \pgfutil@ifnextchar n{\tikz@collect@label@onpath\tikz@hv@lineto}{
+  \pgfutil@ifnextchar n{\tikz@collect@label@onpath\tikz@hv@lineto}{%
   \pgfutil@ifnextchar p{\tikz@collect@pic@onpath\tikz@hv@lineto}%
   {\pgfutil@ifnextchar c{\tikz@collect@coordinate@onpath\tikz@hv@lineto}%
     {\tikz@scan@one@point{\tikz@@hv@lineto}}}}}%
@@ -3225,13 +3225,13 @@
     \let\tikz@collected@onpath=\pgfutil@empty%
     \tikz@edgetoparentcollect%
 }%
-\def\tikz@edgetoparentcollect{
+\def\tikz@edgetoparentcollect{%
   \pgfutil@ifnextchar n{\tikz@collect@label@onpath\tikz@edgetoparentcollect}%
   {%
       \expandafter%
     \endgroup%
     \expandafter\tikz@edgetoparent@rollout\expandafter{\tikz@collected@onpath}%
-  }
+  }%
 }%
 
 \def\tikz@edgetoparent@rollout#1{%
@@ -3861,7 +3861,7 @@
 % A label text always ``ends'' the node.
 %
 \def\tikz@fig ode{%
-  \pgfutil@ifnextchar a\tikz@test@also{
+  \pgfutil@ifnextchar a\tikz@test@also{%
     \pgfutil@ifnextchar f{\tikz@nodes@start}\tikz@normal@fig}}%
 \def\tikz@test@also a{\pgfutil@ifnextchar l\tikz@node@also{\tikz@normal@fig a}}%
 \def\tikz@normal@fig{%
@@ -5261,7 +5261,7 @@
 \def\tikz@scan@absolute#1{%
   \pgfutil@ifnextchar({\tikz@scan@@absolute#1}%)
   {%
-    \advance\tikz@expandcount by -1%
+    \advance\tikz@expandcount by -1
     \ifnum\tikz@expandcount<0\relax%
       \let\pgfutil@next=\tikz@@scangiveup%
     \else%
--- a/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
@@ -2120,10 +2120,15 @@
   \else%
     \tikz@invoke@collected@onpath%
   \fi%
-  \afterassignment\tikz@handle\let\pgf@let@token=%
+  \tikz@scan@@next@command
 }%
+\def\tikz@scan@@next@command{%
+  \afterassignment\tikz@handle\let\pgf@let@token=%
+}
+
 \newcount\tikz@expandcount
 \let\tikz@collected@onpath=\pgfutil@empty%
+\edef\tikz@frozen@relax@token{\ifnum0=0\fi}
 
 % Central dispatcher for commands
 \def\tikz@handle{%
@@ -2224,7 +2229,7 @@
                               \ifx\pgf@let@token:%
                                 \let\pgfutil@next=\tikz@colon@char%
                               \else%
-                                \let\pgfutil@next=\tikz@expand%
+                                \let\pgfutil@next=\tikz@handle@evenmore
                               \fi%
                             \fi%
                           \fi%
@@ -2240,9 +2245,25 @@
       \fi%
     \fi%
   \fi%
-  \ifx\pgfutil@next\tikz@expand\else\tikz@expandcount=100\relax\fi%
   \pgfutil@next%
 }%
+
+% Continued...
+\def\tikz@handle@evenmore{%
+  % if \pgf@let@token is \relax or frozen \relax, skip it and
+  % scan from the next token
+  \ifx\pgf@let@token\relax
+    \let\pgfutil@next=\tikz@scan@@next@command
+  \else
+    \ifx\pgf@let@token\tikz@frozen@relax@token
+      \let\pgfutil@next=\tikz@scan@@next@command
+    \else
+      \let\pgfutil@next=\tikz@expand
+    \fi
+  \fi
+  \ifx\pgfutil@next\tikz@expand\else\tikz@expandcount=100\relax\fi
+  \pgfutil@next
+}
 
 \def\tikz@l@char{%
   \pgfutil@ifnextchar e{\tikz@let@command}{%
--- a/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarycalendar.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarycalendar.code.tex
@@ -393,7 +393,7 @@ execute before day scope={\ifdate{day of month=1}{%
   \pgfutil@next%
 }
 \def\tikz@lib@cal@expand{%
-  \advance\tikz@expandcount by -1%
+  \advance\tikz@expandcount by -1
   \ifnum\tikz@expandcount<0\relax%
     \tikzerror{Giving up on this calendar}%
     \let\pgfutil@next=\tikz@lib@cal@end%


tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
```diff
@@ -2566,7 +2578,7 @@
 
 \def\tikz@skip#1{\tikz@scan@next@command#1}%
 \def\tikz@expand{%
-  \advance\tikz@expandcount by -1%
+  \advance\tikz@expandcount by -1\relax
```
hmenke: Could have also just left the EOL
```suggestion
  \advance\tikz@expandcount by -1
```

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
```diff
@@ -2120,10 +2120,14 @@
   \else%
     \tikz@invoke@collected@onpath%
   \fi%
-  \afterassignment\tikz@handle\let\pgf@let@token=%
+  \tikz@scan@@next@command
 }%
+\def\tikz@scan@@next@command{%
```
hmenke, line 2125: How about your original suggestion?
```suggestion
\def\tikz@handle@next{%
```
If would also be nice to write a short docstring.

hmenke (changes requested):
Let's see what the CI thinks about this.

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
muzimuzhi: Yes, this one is optional. Will remove it since half of the 6 matches returned by `grep -rni '\\advance\\tikz@expandcount' tex` are not ending with `\relax`.

hmenke: Oops, this is gone now, but the `%` at the end of the line is actually wrong.  It should be
```latex
\advance\tikz@expandcount by -1
```

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
muzimuzhi: Original comment: Because it is used by `\tikz@scan@next@command`, how about keeping the current name (which makes it a simplified variant of `\tikz@scan@next@command`), and applying your `\tikz@handle@evenmore` with a docstring?

I forgot that the PR is based on branches hence new commits will automatically added. Sorry I just did what I prefer before receiving your reply.

hmenke: Don't worry.  You can push as many commits as you want.  If it becomes too messy you can always rebase.

hmenke: Sounds good to me.

muzimuzhi:
> Let's see what the CI thinks about this.

CI triggered by (isomorphic) commit https://github.com/hmenke/pgf/commit/cb425a5030fb27021f0da43c52ff4c22cfe0af15 just gets passed.

Discussions about code style here can still continue and I am willing to do a rebasing followed by force pushing at the end.

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
muzimuzhi: > Could have also just left the EOL

Apparently I misunderstood your first comment.

muzimuzhi:
CI / build (dvisvgm) fails when setting up texlive environment, see https://github.com/pgf-tikz/pgf/pull/967/checks?check_run_id=1591472106#step:4:10
``` 
/tmp/texlive/bin/x86_64-linux/tlmgr: TLPDB::from_file could not initialize from: http://mirrors.ibiblio.org/pub/mirrors/CTAN/systems/texlive/tlnet/tlpkg/texlive.tlpdb
/tmp/texlive/bin/x86_64-linux/tlmgr: Maybe the repository setting should be changed.
/tmp/texlive/bin/x86_64-linux/tlmgr: More info: https://tug.org/texlive/acquire.html
Error: Process completed with exit code 1.
```

The CI for the same commit on my forked branch is ok. Guess this failure is caused by network problems.

ilayn:
restarted just to double check

hmenke (approved):

hmenke:
I had to revert this. It broke `circuitikz`.

muzimuzhi:
Could you provide a specific broken example?

I locally compiled the latest manual of `circuitikz` with pgf 3.1.8 (by removing the `-interaction=nonstopmode` in its [`Makefile`](https://github.com/circuitikz/circuitikz/blob/master/Makefile) and then executing `make manual-git`), the compilation passed. The `to path` issue https://github.com/circuitikz/circuitikz/issues/460 you commented also seemed to be irrelevant to this PR.

hmenke:
Sure, sorry. The MWE was only communicated privately by @Rmano.
```latex
\documentclass[border=10pt]{standalone}
\usepackage[siunitx, RPvoltages]{circuitikz}
\begin{document}
\pgfcircversion{} on \pgfversion
\begin{circuitikz}[american]
    \draw (0,0) to [
        R=R,
        v=V,
        i=I,
        f=F,
        ] (3,0);
\end{circuitikz}
\end{document}
```
bisection
```console
$ git bisect start
$ git bisect bad 3.1.8
$ git bisect good 3.1.7a
$ TEXMFHOME=/path/to/pgf git bisect run pdflatex --interaction=nonstopmode test.tex
[...]
79e613ae139f4e16f1f991114fe1a001d0dd9dab is the first bad commit
commit 79e613ae139f4e16f1f991114fe1a001d0dd9dab
Author: muzimuzhi <muzimuzhi@gmail.com>
Date:   Tue Dec 22 05:24:18 2020 +0800

    tikz: handle \relax and frozen \relax on path #966

 tex/generic/pgf/frontendlayer/tikz/tikz.code.tex | 27 +++++++++++++++++++++---
 1 file changed, 24 insertions(+), 3 deletions(-)
bisect run success
```

Rmano:
@muzimuzhi @hmenke
If you check 1.2.7 it does not break anymore, but my solution is more handwaving than other thing. The change that fixed the breakage was https://github.com/circuitikz/circuitikz/commit/14a8f3ab1fe93e1a2c04c2c1409202049e8c7ff4 - adding groups around the node-putting instructions. Using expandable ifs as suggested by @hmenke did not solve the problem.

@muzimuzhi if you check out circuitikz 1.2.6, you can see the thing that puzzled me: if in the MWE you comment out just one of the decorations it will work. Just using the four of them (l= which is the same as giving the parameter to R,  i=, f=, and v=) breaks. Comment one of them and it runs through.

I tried to debug the core problem but it exceeds my TeX skills, and found almost by chance that an extra scope solved (or masked?) the problem. 

I am a bit busy today but I can help debugging.

hmenke:
@Rmano No, I don't want you to add workarounds for PGF regressions.  Please always report them, because you are most likely not the only one who is affected.

Rmano:
Ok, let's see if I can understand *why* it broke, will remove the workaround in the next minor. My hunch, probably wrong, is that it was hitting some expansion limit, like the famous ` \tikz \node{\boldmath $x$};` (see https://tex.stackexchange.com/questions/487777/why-boldmath-fails-in-a-tikz-node); due to that hunch I tried to add groups --- and it worked.

muzimuzhi:
> The MWE was only communicated privately by @Rmano.

@hmenke If you referred to the chats on TeX-SX, they're public. See https://chat.stackexchange.com/transcript/message/56560034#56560034.

I think the real cause is a hidden problem that `\tikz@expandcount` is not reset in cases handled by `\tikz@handle`, which was made much more easier to trigger by introducing `\tikz@handle@evenmore`. See a possible fix in https://github.com/muzimuzhi/pgf/commit/0f0b925dadc588741f4522f1d0369a1b2d1a9083, which still contains some bit of debugging code. (And I'm not sure if the beginning `\let\pgfutil@next\tikz@expand` in `\tikz@handle` is needed, though CI didn't complain about the deletion.)

The (expected) logic in `\tikz@handle` is,
 - if the collected token is handled, then reset `\tikz@expandcount` to 100 and continue with specific commands; 
 - otherwise, reduce that count by 1, put back the collected token and do one-step expansion on it, re-collect it, and finally try `\tikz@handle` on it again.
 - if the count is lower than zero after reducing, throw an error.

By introducing `\tikz@handle@more`, the cases handled by `\tikz@handle` never reset the count. This has already buried a small hidden danger. By introducing `\tikz@handle@evenmore`, almost every handled cases, except for the two newly added ones, will not reset the count. The difference gets accumulated by occasional one-step expansions, and eventually explode in a complex (as of its internal expansions, not the user input) enough `circuitikz` example. Since the change to`\tikz@expandcount` is local, adding an extra group to some part of the tikz path happens to solve the problem.

Some intermediate results that helped me when digging the problem:
 - `circuitikz` does not generates either forms of `\relax`.
 - Adding new `\ifx...\else...\fi`s in `\tikz@handle@more` is ok, but introducing `\tikz@handle@evenmore` is buggy.
 - The value of `\tikz@expandcount` at the beginning of `\tikz@expand` is frequently reset before, but never/seldom gets reset when this PR is applied.

hmenke:
@muzimuzhi If you want we can try again with your new patch.  However, after thinking a bit more about this, encountering `\relax` or frozen `\relax` on the path is usually a sign of a different problem in the code that the user wrote.  I'm not so sure whether we should really handle it in this way.  It's probably better to error out with a more specific error in these cases.  Currently it throws `Did you forget a semicolon?` which often is not the case at all.

Rmano:
So, if I have understood correctly: that mechanism you describe is a safety mechanism to avoid infinity recursion --- but the counter was not reset so that using more than three "decorations" hit the limit value. So, adding the extra grouping made the changes to the counter local, and so it did not trigger the limit. Am I correct?

hmenke:
@Rmano That describes it pretty accurately.  You skipped the point where I fucked it up, though.

Rmano:
> @Rmano That describes it pretty accurately. You skipped the point where I fucked it up, though.

:-; you can try to blame somebody else like in the chat, seems quite standard :-P

muzimuzhi:
Two discussions are mixed up. One for questioning why pgf 3.1.8 + circuitikz 1.2.6 is broken for some example, and another one for whether tikz should silently skipping `\relax` and frozen `\relax` in path handling.

The first one is what @Rmano is interested/involved in, if not the only one. And its cause is already uncovered. There is nothing wrong in circuitikz 1.2.6, but something wrong in pgf, especially 3.1.8. And that cause could have nothing to do with current PR if we adjusted the way to implement, for example the one originally posted in https://github.com/pgf-tikz/pgf/issues/966#issuecomment-749154057. Relax, it's ok if the second one didn't attract you.

The second one is what this PR wanted to address.

@hmenke My original purpose was to i) make the similar logic written in `\foreach ... {\draw ...;}` and `\draw foreach ...;` either both work or both not work, and ii) make the normal texnique `\if(num|dim)\x=\y\relax` works in foreach path operation, because as I could imagine, explaining frozen `\relax` or adding `\dimexpr` to users is more difficult than just `\relax`. The remained question is, as you pointed out, whether auto ignoring `\relax` will cover any real problems, rather than ease minor ones. This also relates to the error message concern. Another consideration is, should the complex/nested foreach path operation (not the `\foreach` command used outside tikz path) be encouraged or discouraged. (I may have encountered the `\relax` problem before, and resolved by rewriting codes to `\foreach ... {\draw ...}`.)

I don't have a strong opinion on both of them (about second one).

muzimuzhi:
> However, after thinking a bit more about this, encountering `\relax` or frozen `\relax` on the path is usually a sign of a different problem in the code that the user wrote.

In short, need some examples.

hmenke:
I think we can handle `\relax` in a reasonable way by stepping over it without removing it.  It just came to my mind that with the original patch `\relax` was actually discarded.  That might or might not be a problem.  From my point of view it's also really hard to argue what's the correct behavior here, because we're somewhere in the middle of a complex expansion chain.  Another big problem here is actually `\expandafter\tikz@scan@next@command\pgf@let@token` which will happily expand `\protected` material.  Maybe you could open a new PR with your revised patch and then we can work on it together.

muzimuzhi:
> Another big problem here is actually `\expandafter\tikz@scan@next@command\pgf@let@token` which will happily expand `\protected` material.

Aha, seems another call for latex3 expandable macros `\token_if_xxx` implemented in pgf. Not sure if I correctly got your point. To solve the problem piece by piece, firstly I opened PR #970 that solely addresses the `\tikz@expandcount` issue.

PR #970, opened 2020-12-28
tikz: reset \tikz@expandcount more frequent #969

### LINKED ISSUE #969: tikz should reset \tikz@expandcount more frequent
**Brief outline of the bug**

As discussed in https://github.com/pgf-tikz/pgf/pull/967#issuecomment-751710949, currently tikz never resets `\tikz@expandcount` when a token handled in `\tikz@handle` is encountered on a path. This will raise error `Giving up on this path. Did you forget a semicolon?` when too many steps of expansions are executed when scanning a path.

```latex
\documentclass{article}
\usepackage{tikz}

\makeatletter
% Token "r" is handled in \tikz@handle. 
% To show the problem, redefine its handler.
\let\tikz@rect\tikz@scan@next@command

\def\oneHundredExpansions{}
\def\oneExpansion{r}

\foreach \i in {1,...,101} { % 100 passes, 101 throws error
  \g@addto@macro\oneHundredExpansions{\oneExpansion}
}
\makeatother

\begin{document}
\begin{tikzpicture}
  % One handled "r" is encountered every two expansions, but the 
  % \tikz@expandcount is never reset until the 101st expansion throws an error.
  \draw (0,0) -- (1,1) \oneHundredExpansions;
\end{tikzpicture}
\end{document}
```

```
! Package tikz Error: Giving up on this path. Did you forget a semicolon?.
```

@Rmano
Note that in pgf 3.1.8 and 3.1.8a, tokens handled by `\tikz@handler@more` will not reset the count as well. In the above example, they are `(` in `(0, 0)`, `-` in `--`, and `(` in `(1, 1)`, hence `\foreach \i in {1,...,98}` is enough to trigger the error. See a therefore broken `circuitikz` example in https://github.com/pgf-tikz/pgf/pull/967#issuecomment-751616162.

muzimuzhi:
This makes `\tikz@expandcount` get reset every time a token is handled
in tikz path scanning.

Fixes #969

**Checklist**

### DIFF
--- a/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarycalendar.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarycalendar.code.tex
@@ -370,28 +370,13 @@ execute before day scope={\ifdate{day of month=1}{%
 }
 
 \def\tikz@lib@cal@handle{%
-  \let\pgfutil@next=\tikz@lib@cal@expand%
-  \ifx\pgf@let@token;%
-    \let\pgfutil@next=\tikz@lib@cal@stop%
-  \else%
-    \ifx\pgf@let@token(%)
-      \let\pgfutil@next=\tikz@lib@cal@name%
-    \else%
-      \ifx\pgf@let@token a%
-        \let\pgfutil@next=\tikz@lib@cal@at%
-      \else%
-        \ifx\pgf@let@token[%
-          \let\pgfutil@next=\tikz@lib@cal@option%
-        \else%
-          \ifx\pgf@let@token i%
-            \let\pgfutil@next=\tikz@lib@cal@if%
-          \fi%
-        \fi%
-      \fi%
-    \fi%
-  \fi%
-  \pgfutil@ifx\pgfutil@next\tikz@lib@cal@expand{}{\tikz@resetexpandcount}%
-  \pgfutil@next%
+  \pgfutil@switch\pgfutil@ifx\pgf@let@token{%
+    {;}{\tikz@lib@cal@stop}%
+    {(}{\tikz@lib@cal@name}%)
+    {a}{\tikz@lib@cal@at}%
+    {[}{\tikz@lib@cal@option}%]
+    {i}{\tikz@lib@cal@if}%
+  }{}{\tikz@resetexpandcount\tikz@lib@cal@expand}%
 }
 \def\tikz@lib@cal@expand{%
   \advance\tikz@expandcount by -1
--- a/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarycalendar.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarycalendar.code.tex
@@ -358,7 +358,7 @@ execute before day scope={\ifdate{day of month=1}{%
 \def\tikz@lib@cal@calendar{%
   \begingroup%
     \let\tikz@lib@cal@ifs=\pgfutil@empty%
-    \tikz@expandcount=1000\relax%
+    \tikz@resetexpandcount
     \tikzset{name=,at={(0,0)}}%
     \let\%=\pgfcalendarshorthand%
     \tikzset{every calendar/.try}%
@@ -390,6 +390,7 @@ execute before day scope={\ifdate{day of month=1}{%
       \fi%
     \fi%
   \fi%
+  \pgfutil@ifx\pgfutil@next\tikz@lib@cal@expand{}{\tikz@resetexpandcount}%
   \pgfutil@next%
 }
 \def\tikz@lib@cal@expand{%
--- a/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
@@ -2173,6 +2173,7 @@
       \fi%
     \fi%
   \fi%
+  \ifx\pgfutil@next\tikz@handle@more\else\tikz@expandcount=100\relax\fi
   \pgfutil@next%
 }%
 
--- a/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
@@ -2105,7 +2105,7 @@
     \tikz@snakedfalse%
     \tikz@decoratepathfalse%
     \tikz@node@is@a@labelfalse%
-    \tikz@expandcount=100\relax%
+    \tikz@resetexpandcount
     \pgf@path@lastx=0pt%
     \pgf@path@lasty=0pt%
     \tikz@lastx=0pt%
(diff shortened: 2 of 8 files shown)


tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
```diff
@@ -2173,6 +2173,7 @@
       \fi%
     \fi%
   \fi%
+  \ifx\pgfutil@next\tikz@handle@more\else\tikz@expandcount=100\relax\fi
```
hmenke: I've changed `\tikz@handle@more` to `\tikz@expand` here.  I don't see why the count should be reset in the case of `\tikz@handle@more`.

tex/generic/pgf/frontendlayer/tikz/libraries/tikzlibrarycalendar.code.tex
```diff
@@ -358,7 +358,7 @@
 \def\tikz@lib@cal@calendar{%
   \begingroup%
     \let\tikz@lib@cal@ifs=\pgfutil@empty%
-    \tikz@expandcount=1000\relax%
```
hmenke, line 361: 1000 seemed really excessive.  It was probably set to 1000 because it would never be reset.

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
muzimuzhi: No it should be `\tikz@handle@more` here. At the end of `\tikz@handle`, `\pgfutil@next` is either one of the handlers or `\tikz@handle@more`, but is always not `\tikz@expand`. The logic in my commit was to reset the count when `\pgf@let@token` _is_ handled in `\tikz@handle`. Using `\pgfutil@ifx\pgfutil@next\tikz@expand` will reset the count even when `\pgf@let@token` is eventually not handled in `\tikz@handle@more`.

muzimuzhi:
appreciate the cleaning up

hmenke:
@muzimuzhi What do you think about https://github.com/pgf-tikz/pgf/pull/970/commits/a9a0f97553671b5ade2260cf6619db5e133b93e4?

muzimuzhi:
very well, even impressive (except for the lua emulated `\pgfsys@strcmp` because I know little about luatex)

muzimuzhi:
@hmenke Are there any concerns that blocked this PR to be merged? Considering its current title, I think it has included well enough changes/improvements.

hmenke:
Should we discuss again the “`\relax` on a path” issue?  We could add another case distinction like
```latex
{\relax}{\relax\tikz@scan@next@command}%
```
Which would reinsert `\relax` and continue scanning.  Does that solve the `\ifnum\x=\y\relax` thing?

muzimuzhi:
Ah, I mistakenly though the `\relax` would be handled in another PR.

Yes I guess that would solve the `\ifnum...` thing, and will  test it soon. Noticed `\par` is already handled as `{\par}{\tikz@scan@next@command}`, using `{\relax}{\relax\tikz@scan@next@command}` looks convincing.

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
```diff
@@ -2123,125 +2123,43 @@
...
+    {\pgfextra}{\tikz@extra}%
+    {\foreach}{\tikz@foreach}%
+    {f}{\tikz@fchar}%
+    {\pgf@stop}{\relax}%
+    {\par}{\tikz@scan@next@command}%
+    {d}{\tikz@decoration}%
+    {l}{\tikz@l@char}%
+    {:}{\tikz@colon@char}%
+    {\relax}{\relax\tikz@scan@next@command}%
+    {\tikz@frozen@relax@token}{\tikz@frozen@relax@token\tikz@scan@next@command}%
```
hmenke: Do you think we should handle frozen relax in this way?  This only happens when TeX failed to read a number which means that the statement that wanted to use the number becomes nonsensical.  I think this will be even more confusing for users than `Giving up on this path. Did you forget a semicolon?`

muzimuzhi: Reasonable. Having `\relax` handled is enough for inputs like `\draw (0,0) \ifnum0=0\relax\fi -- (1,1);` with "standard" texniques applied.  I will remove the frozen `\relax` part. TeX is just too dynamic/free to not define DSLs one after another but too weak to assert user input and catch errors.

hmenke:
TeX Live failed to update.  Let's see if rerunning helps.

hmenke (approved):

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
```diff
@@ -2123,125 +2123,42 @@
...
+    {\pgfextra}{\tikz@extra}%
+    {\foreach}{\tikz@foreach}%
+    {f}{\tikz@fchar}%
+    {\pgf@stop}{\relax}%
+    {\par}{\tikz@scan@next@command}%
+    {d}{\tikz@decoration}%
+    {l}{\tikz@l@char}%
+    {:}{\tikz@colon@char}%
+    {\relax}{\relax\tikz@scan@next@command}%
+  }{}{\tikz@resetexpandcount\tikz@expand}%
```
muzimuzhi, line 2161: This is buggy. The count should be reset in case of match, rather than not match. Current code will result in endless loop.

Moving `\tikz@resetexpandcount` to `#4` of `\pgfutil@switch` doesn't work either because then e.g. `\tikz@movetoabs\tikz@resetexpandcount` is left in input stream, but we need them in reversed order, <strike>at least the expansion order</strike>.

muzimuzhi: A possible fix https://github.com/muzimuzhi/pgf/compare/ea44e24e5f721b9b4a00e26f8291ebfc76522ff6...6ffd8f92cbd84eeaaadee161f5d7a59fc9e000a3, note this makes the semantic of `\pgfutil@switch` un-conventional.

hmenke: I think we have to go with
```latex
  % ..
  {:}{\let\pgf@next\tikz@colon@char}%
  {\relax}{\relax\let\pgf@next\tikz@scan@next@command}%
}{\tikz@resetexpandcount\pgf@next}{\tikz@expand}%
```
Sorry, but I want to avoid the changed semantics of `\pgfutil@switch`. In fact I was thinking about moving this to a separate `pgfexpl` module which can serve as a drop-in replacement for `expl3` in case we want to switch to that in the future.  That would make migration as easy as doing a simple search-and-replace.

muzimuzhi: That's rather reasonable (and exciting for the `pgfexpl` part).  Could I understand it as you have assigned yourself to the fixing and new `pgfexpl` tasks? 😄

muzimuzhi: Ah, I can do the part of fixing uses of `\pgfutil@switch`. And leave the part of moving and renaming `\pgfutil@switch` (and possibly `pgfint.code.tex`) to you.

_Update:_ <strike>Done in https://github.com/muzimuzhi/pgf/commit/590c5e92d5cc21c0cb8ef3ab551131cb70cc7ef1 </strike>. See #981.

muzimuzhi: Found it's `\token_case_meaning:Nn` that `\pgfutil@switch\pgfutil@ifx` imitates. I should read some `expl3` codes ...

PR #994, opened 2021-02-22
documentation fix in introduction: syntax is similar to METAPOST not METAFONT

itmm:

corrected documentation

**Checklist**

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-introduction.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-introduction.tex
@@ -95,7 +95,7 @@ beginning a path at the first corner of the triangle, one for extending the
 path to the second corner, one for going to the third, one for closing the
 path, and one for actually painting the triangle (as opposed to filling it).
 With the \tikzname\ frontend all this boils down to a single simple
-\textsc{metafont}-like command:
+\textsc{metapost}-like command:
 %
 \begin{verbatim}
 \draw (0,0) -- (1,0) -- (1,1) -- cycle;
@@ -103,7 +103,7 @@ With the \tikzname\ frontend all this boils down to a single simple
 
 In practice, \tikzname\ is the only ``serious'' frontend for \pgfname. It gives
 you access to all features of \pgfname, but it is intended to be easy to use.
-The syntax is a mixture of \textsc{metafont} and \textsc{pstricks} and some
+The syntax is a mixture of \textsc{metapost} and \textsc{pstricks} and some
 ideas of myself. There are other frontends besides \tikzname, but they are intended
 more as ``technology studies'' and less as serious alternatives to
 \tikzname. In particular, the |pgfpict2e| frontend   reimplements the standard


muzimuzhi (approved):

muzimuzhi:
Thanks.

hmenke:
That's actually wrong.  The MetaPost syntax is inherited from the MetaFont syntax.  Also I find it very questionable to "correct" the original author's remarks on the design.  If Till was inspired by the MetaFont syntax, then that's it.  Please revert this.

itmm:
The rest of the introduction only mentions MetaPost, not MetaFont. I do not mind if you revert the change, but I got confused while reading it and thought, the result would be clearer, if you refer to the already mentioned package.

muzimuzhi:
An quick investigation in pgfmanual v3.1.8b shows:

| Words | No. | pgfmanual info | Contexts | Source pos. (WIP) | Notes |
| -------------- | ---- | ---------------------------------------------------------- | ------------------------------------------------------------ | --------------------- | ------- |
| `metafont`     | 1    | Sec. 1.1.1 "The Layers Below TikZ", p. 27                  | With the TikZ frontend all this boils down to a single simple **metafont**-like command:<br />`\draw (0,0) -- (1,0) -- (1,1) -- cycle;` |                       | [1, 3]           |
|                | 2    | ___, p. 28                                                 | The syntax is a mixture of **metafont** and pstricks and some ideas of myself. |                       | [1]              |
|                | 3    | Sec. 11 "Design Principles", p. 124                        | The basic command names (of TikZ) and the notion of path operations is taken from **metafont**; ... |                       |                  |
|                | 4    | Sec. 15.3.2 "Graphic Parameters: Dash Pattern", p. 175     | `/tikz/dash pattern=<dash pattern>`<br />Sets the dashing pattern. The syntax is the same as in **metafont**. |                       | [2]              |
|                | 5    | Sec. 25 "Transformations", p. 374                          | pgf has a powerful transformation mechanism that is similar to the transformation capabilities of **metafont**. |                       |                  |
| `metapost`     | 1, 2 | Sec. 1.2 "Comparison with Other Graphics Packages", p. 28  | The **metapost** program is a powerful alternative to TikZ. It used to be an external program, which entailed a bunch of problems, but in LuaTeX it is now built in. An obstacle with **metapost** is the inclusion of labels. |                       |                  |
|                | 3    | Sec. 11.2 "Special Syntax For Path Specifications", p. 124 | TikZ makes it easy to specify paths, partly using the syntax of **metapost**. For example, to specify a triangular path you use<br />`(5pt,0pt) -- (0pt,0pt) -- (0pt,5pt) -- cycle` |                       | [3]              |

[1] These two relate to current PR and to common syntax of both `metafont` and `metapost`.
[2] This one should be `metapost` because `metafont` provides no syntax for dash pattern.
[3] Similar tikz syntax is related once to each of `metapost` and `metafont`. Should it stick to `metafont`?

@itmm Since the original author is inactive and we could only inter his intention, it seems most referring to `metafont` is ok and the proposed change in current PR is not necessary. Therefore I will revert this PR tonight, in case the discussion about the last two table notes will result in a different decision.

@hmenke Sorry for my hurry approval and merging. And what's your opinion on the last two notes?

muzimuzhi:
This is reverted in https://github.com/pgf-tikz/pgf/commit/a02c45a62ccfdcf57a5e3a0fd0b2042aeee1a879, currently on `staging` branch.

(This seems to not change the state of current PR, which is "merged". Should I have opened a new PR to do the reverting?)

hmenke:
That's fine.  I just applied the wont-fix label.

PR #890, opened 2020-06-29
gs-9.53 transparency (fixes #862); blend mode; transparency groups 

agrahn:
**1.** GS 9.52 introduced new transparency operators

````
.setfillconstantalpha
.setstrokeconstantalpha
````

Cf. issue #862.

They are supposed to replace `.setopacityalpha` and allow opacity to be set individually for stroking and filling operations. `.setopacityalpha` will be deprecated with the upcoming version 9.53. This PR makes use of the new operators, but keeps `.setopacityalpha` as a fallback.

**2.** This PR also implements Transparency Groups and Blend Modes.

----

The code listing below contains examples from the PGF manual. Typeset with:
````
latex example.tex
dvips example
ps2pdf -dALLOWPSTRANSPARENCY example.ps
````
Outputs from `ps2pdf` and `pdftex` for comparison:
[transparencyExamples-dvips.pdf](https://github.com/pgf-tikz/pgf/files/4846538/transparencyExamples-dvips.pdf)
[transparencyExamples-pdftex.pdf](https://github.com/pgf-tikz/pgf/files/4846539/transparencyExamples-pdftex.pdf)

(Poppler-based viewers don't render correctly the knockout example in the `pdftex` output, but MuPDF and AcroReader do.)

````latex
\documentclass{article}
\pagestyle{empty}

\usepackage{tikz}
\usetikzlibrary {patterns,shapes.symbols}

\begin{document}
  \begin{tikzpicture}[thick,fill opacity=0.5]
    \filldraw[fill=red] (0:1cm) circle (12mm);
    \filldraw[fill=green] (120:1cm) circle (12mm);
    \filldraw[fill=blue] (-120:1cm) circle (12mm);
  \end{tikzpicture}
%
  \begin{tikzpicture}[every node/.style={fill,draw}]
    \draw[line width=2mm,blue!50,line cap=round] (0,0) grid (3,2);
    \node[opacity=0.5] at (1.5,2) {Upper node};
    \node[draw opacity=0.8,fill opacity=0.2,text opacity=1] at (1.5,0) {Lower node};
  \end{tikzpicture}
%
  \begin{tikzpicture}[fill opacity=0.5]
    \fill[red] (0,0) circle (1);
    \fill[red] (1,0) circle (1);
  \end{tikzpicture}

  transparency group:
  \begin{tikzpicture}[opacity=0.5]
    \begin{scope}[transparency group]
      \draw [line width=5mm] (0,0) -- (2,2);
[...]

PR #996, opened 2021-02-27
dvips: fix displacement after blend group, #995

### LINKED ISSUE #995: dvips backend: blend group displaces text and graphics
**Brief outline of the bug**
When converting the below example using `latex` and `dvips`, text and graphics are displaced in the resulting PS file. The "X" is supposed to be placed inside the circle but appears outside. The issue goes away if `blend group = soft light` is removed from the `scope` environment.
`pdflatex`, in contrast, converts the example correctly. 

This is a reduced version of the [Venn diagram example](https://texample.net/tikz/examples/venn).
```latex
\documentclass{article}
\usepackage{tikz}
\begin{document}
\begin{tikzpicture}
  \begin{scope}[blend group = soft light]
    \fill[red!30!white] (90:1.2) circle (2);
  \end{scope}
  \node at (90:2) {X};
\end{tikzpicture}
\end{document}
```
venn

muzimuzhi:

From the example in #995, I found that using dvi-ps-pdf chain, node text after a blend group is displaced, and the displacement distance is proportional to the  size of the picture (actually the x-size for x < 0pt) just before the node was encountered.

According to the similar treatment in `\endpgftransparencygroup` (see source code snippet below), it seems a `\hskip-\pgf@picminx` is missing in `\pgfsys@transparencygroupfrombox` for dvips driver.
https://github.com/pgf-tikz/pgf/blob/b1d133cea43c256112f383906a39b10232030137/tex/generic/pgf/basiclayer/pgfcoretransparency.code.tex#L300-L304

Misc. info: The blend group support for dvips was added in #890.

Fixes #995

**Checklist**

### DIFF
--- a/tex/generic/pgf/systemlayer/pgfsys-dvips.def
+++ b/tex/generic/pgf/systemlayer/pgfsys-dvips.def
@@ -567,7 +567,7 @@
   \wd#1=0pt
   \ht#1=0pt
   \dp#1=0pt
-  \hskip\pgf@picminx\box#1%
+  \hskip\pgf@picminx\box#1\hskip-\pgf@picminx
   \pgfsys@outerinvoke{%
     /pgffoa pgf@sys@foa def /pgfsoa pgf@sys@soa def
     grestore


muzimuzhi:
@hmenke Thanks. I should have explicitly request reviews from both you and @agrahn. But taking into account that you are so responsive ...

PR #1018, opened 2021-06-01
Ensure \tracinglostchars<3 in \pgf@picture

PhelypeOleinik:

Setting `\tracinglostchars=3` (or 2 for the faint of heart) is a must-have when writing documents to avoid missing characters slipping by, but it may break TikZ because it allows more or less arbitrary things to be written with `\nullfont`.  This PR ensures that `\tracinglostchars` is at most 2 when a `\pgf@picture` starts.

**Checklist**

### DIFF
--- a/tex/generic/pgf/basiclayer/pgfcorescopes.code.tex
+++ b/tex/generic/pgf/basiclayer/pgfcorescopes.code.tex
@@ -242,6 +242,8 @@
           \let\addtolength=\pgf@addtolength%
           \let\selectfont=\pgf@selectfont%
           \nullfont\spaceskip0pt\xspaceskip0pt%
+          \tracinglostchars=%
+            \ifnum\tracinglostchars>2 2\else\tracinglostchars\fi
           \setbox\pgf@layerbox@main\hbox to0pt\bgroup%
             \begingroup%
   }


hmenke:
Unfortunately this has the potential of breaking lots of old (arguably defective) documents. I'm not sure how to proceed.

PhelypeOleinik:
The proposed implementation only changes the value of `\tracinglostchars` when it was equal to or greater than 3 outside of the picture.  This means that the only incompatible change will be in case the document had set `\tracinglostchars=3` outside of the picture and relied on that value inside of the picture.

Unless I'm missing some edge-case, this should not be a big issue because the different behaviour for `\tracinglostchars=3` was only implemented last year (thus first available in TL 21).

hmenke:
Oh, I see. I got it the wrong way around. At first I though this PR would set `\tracinglostchars=3` inside the picture.
For a moment I was wondering whether to just keep `\tracinglostchars=3` inside the picture, because often missing characters are [bugs in PGF](https://github.com/pgf-tikz/pgf/issues?q=is%3Aissue+missing+character), but then again, do I really want so many reports? Probably not.

PhelypeOleinik:
Thanks for merging!

Yeah, setting `\tracinglostchars=3` there would be a bold move ;-)

PR #1026, opened 2021-06-18
Refine error recovery after missing semicolon

### LINKED ISSUE #1025: error recovery after missing semicolon
**Brief outline of the proposed feature**

If the user omits a semicolon then tikz issues a quite reasonable error

`! Package tikz Error: Giving up on this path. Did you forget a semicolon?.`

But if you scroll past that it often ends up in a tight infinite loop so manages to run forever without filling any stack and stopping.

This is particularly troubling with systems like latexmk (and its deployment in Overleaf) as the user never sees the terminal and as the job is killed via a timout the log never gets written properly and the user is never shown the error.

I suggest below making a hard stop after the error message, this works for the Overleaf use case but maybe a bit agressive but perhaps after the error it would be possible to do a quick test if you were in a probable loop and only using `\@@end` in that case. But otherwise I think always doing a hard stop is safer than looping.

I give a possible code change below but you can see the effect this has on the overleaf UI 

current timeout message:

https://www.overleaf.com/read/jwbvcmsgxhzq

Modified example showing missing semicolon error

https://www.overleaf.com/read/zxzfhxjspvcq

```latex
\documentclass{article}

\usepackage{tikz}
\makeatletter
\def\tikz@expand{%
  \advance\tikz@expandcount by -1
  \ifnum\tikz@expandcount<0\relax%
    \tikzerror{Giving up on this path. Did you forget a semicolon?}%
%    \let\pgfutil@next=\tikz@finish%
    \def\pgfutil@next{\@@end}%
  \else%
    \let\pgfutil@next=\tikz@@expand
  \fi%
  \pgfutil@next}%
\makeatother

\begin{document}

\begin{tikzpicture}
\node at (0,0) {A}
\node at (1,1) {B};
\end{tikzpicture}
 
\end{document}

```

muzimuzhi:

This PR tries to make https://github.com/hmenke/pgf/commit/8a6c14c9789e8bdc687506afa7b6723bb8249717 pass the <s>CI</s> manual building. Currently, two compatibility issues are already handled, one with `chains` lib and another with `scopes` lib.

I made it as a PR so other `pgf-tikz` members could directly add commits to fix remaining problems.

Fixes #1025

An example for testing
```tex
\documentclass{article}
\usepackage{tikz}
\usetikzlibrary{chains, scopes}

\begin{document}
% Test with `chains` lib, when \pgf@let@token contains an \else
\begin{tikzpicture}[
    start chain,
    every node/.style={on chain, join},
  ]
  
  \node (a)     {X};
  \node (b)     {Y};
\end{tikzpicture}

% Test with `scopes` lib
\begin{tikzpicture}
  \node[red] (existing) at (2,0) {Y};
  
  {[start chain]
    \node [on chain,join] {X};
    \chainin (existing) [join];
  }
\end{tikzpicture}

% Test error recovery after missing semicolon
\begin{tikzpicture}
  \draw (0,0) -- +(1,1) %;
  \draw (0,1) -- +(1,-1);
\end{tikzpicture}

\end{document}
```

**Checklist**

### DIFF
--- a/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
@@ -2486,12 +2486,27 @@
 \def\tikz@expand{%
   \advance\tikz@expandcount by -1
   \ifnum\tikz@expandcount<0\relax%
+    \expandafter\pgfutil@firstoftwo
+  \else
+    \expandafter\pgfutil@secondoftwo
+  \fi
+  {%
     \tikzerror{Giving up on this path. Did you forget a semicolon?}%
-    \let\pgfutil@next=\tikz@finish%
-  \else%
-    \let\pgfutil@next=\tikz@@expand
-  \fi%
-  \pgfutil@next}%
+    % since the last token caused an error we should reinsert it and therefore save it
+    \global\let\tikz@expand@last@token=\pgf@let@token
+    \tikz@finish%
+    %
+    % To be combatible with `scopes` lib, which uses a redefined 
+    % \tikz@lib@scope@check to check the next token, the reinsertion is done
+    % here, not at the end of (every) \tikz@finish.
+    %
+    \expandafter\let\expandafter\tikz@expand@last@token@\csname tikz@expand@last@token\endcsname
+    \global\let\tikz@expand@last@token=\relax
+    \tikz@expand@last@token@
+  }{%
+    \tikz@@expand
+  }%
+}
 
 \def\tikz@@expand{%
   \expandafter\tikz@scan@next@command\pgf@let@token}%


hmenke:
I've pushed the typo fixes independently. Can you rebase and squash the rest? I think it should be all in a single commit.

muzimuzhi:
@hmenke Done. Looks like I cannot make you as [co-author](https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#required-co-author-information), since you "requires that any commit attributed to them is signed".

hmenke:
There is still a problem with `scopes` but that is just loosely related to the issue at hand.
```latex
\documentclass{article}
\usepackage{tikz}
\usetikzlibrary{chains, scopes}
\begin{document}
\begin{tikzpicture}
  \node[red] (existing) at (2,0) {Y};
  
  {[start chain]
    \node [on chain,join] {X};
    \chainin (existing) [join] % oops
  }
\end{tikzpicture}
\end{document}
```
In this case the parser sees `}` on the path, which is handled in `\tikz@handle`

https://github.com/pgf-tikz/pgf/blob/e9c22dc9fe48f975b7fdb32181f03090b3747499/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex#L2144

which then misses inserting `\egroup` and causes a spew of unrelated errors culminating in `! Infinite glue shrinkage found in a paragraph.`

muzimuzhi:
Guess there are more cases in the form of `{ ... <tikz path without semicolon> }` with `scopes`. 

To fully complete the current task, `tikz` need to know which kinds of next token are permissible more specifically, rather than leave the work to `\tikz@handle`. For example, right after `\chainin (<node name>)`, only `[` and `;` are permissible; and after `\chainin (<node name>) [<options>]`, only `;` is permissible (maybe another `[`). This may lead to a clearer and stricter description of `tikz`'s syntax.

muzimuzhi:
More troublesome is, in the case of `{ ... \draw ... [...] <a> [...] <b> }` with `scopes`, a missing `;` can be inserted at either `<a>` and `<b>`. So compared to user's intension, the behavior of any reinsertion logic might be mistakes. This requires the logic itself be documented.

hmenke:
I don't think `scopes` is that widely used, so from my point of view it is fine to just keep the obscure errors and wait for someone to complain about it.

PR #1029, opened 2021-07-11
To allow pgfpages with TRT page direction

### LINKED ISSUE #911: Adaptation of pgfpages for \pagedir TRT (lualatex)
**Brief outline of the proposed feature**

With lualatex and `\pagedir TRT` layouts are shifted out of page this is due to that  the reference point of pages  is at lower left corner but with `\pagedir TRT`   the reference point of pages change to  lower right  

A solution is to reverse  x coordinate in `\pgfpoint` command and this give correct result see
https://tex.stackexchange.com/a/559842/54817

Maybe this can be added to `pgfpages` like this 

```latex
\newif\ifpgfpagedir@trt
\pgfpagedir@trtfalse

\newcommand{\pgfpagedirtrt}{\pgfpagedir@trttrue}

% Perhaps we should add \ifluatex 
% \newcommand{\pgfpagedirtrt}{\ifluatex\pgfpagedir@trttrue\else\relax\fi} 

\let\pgfpoint@TLT\pgfpoint
\def\pgfpoint@TRT#1#2{\pgfpoint@TLT{-1*#1}{#2}}
```
Then for every `pgfpages` commands containing `\pgfpoint` we need to add 

```latex
\ifpgfpagedirtrt\let\pgfpoint\pgfpoint@TRT\fi
```

**Usage example**

```latex
% lualatex 
\documentclass{article}
\usepackage[showframe]{geometry}
\usepackage{pgfpages}

\pgfpagedirtrt <-- % to reverse x coordinate in \pgfpoint inside pgfpages commands 

\pgfpagesuselayout{4 on 1}[a4paper,landscape]

\begin{document}

\pagedir TRT \bodydir TRT \pardir TRT 

\Huge 

Page 1
  
\newpage

Page 2

\newpage

Page 3

\newpage

Page 4

\end{document}
```

seloumi:

Fixes #911 and allow using `pgfpages` in case of TRT page direction (lualatex)
  just with `\def\pgfpagespoint#1#2{\pgfpoint{-#1}{#2}}`

**Checklist**

- [ ] Documentation changes are licensed under [FDLv1.2][FDL]

### DIFF
--- a/tex/latex/pgf/utilities/pgfpages.sty
+++ b/tex/latex/pgf/utilities/pgfpages.sty
@@ -34,7 +34,27 @@
 \pgfphysicalpageemptytrue
 \pgf@holdingphysicalpagefalse
 
-\let\pgfpagespoint\pgfpoint
+\def\pgfpagespoint{%
+  % Delay the definition of \pgfpagespoint in case the \pagedir is set at the
+  % beginning of document, but still assume that the \pagedir is constant 
+  % throughout the document.
+  \pgfpages@test@pagedir\pgfpagespoint}
+
+\def\pgfpages@test@pagedir{%
+  \let\pgfpagespoint\pgfpoint
+  \ifdefined\luatexversion
+    % \pagedir returns a three-char string and \pagedirection returns an int
+    \ifcase\pagedirection
+        % 0, TLT (latin)
+    \or % 1, TRT (arabic)
+      \def\pgfpagespoint##1##2{\pgfpoint{##1}{##2}\pgf@x=-\pgf@x}%
+    \or % 2, LTL (mongolian)
+      % ... need any transformations?
+    \or % 3, RTT (cjk)
+      % ... need \pgf@x=-\pgf@x ?
+    \fi
+  \fi
+}
 
 % Define a layout
 %
@@ -1137,4 +1157,3 @@
     \pgfshipoutphysicalpage
   \fi
 }
-
--- a/tex/latex/pgf/utilities/pgfpages.sty
+++ b/tex/latex/pgf/utilities/pgfpages.sty
@@ -34,6 +34,7 @@
 \pgfphysicalpageemptytrue
 \pgf@holdingphysicalpagefalse
 
+\let\pgfpagespoint\pgfpoint
 
 % Define a layout
 %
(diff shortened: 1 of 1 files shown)


seloumi:
**Example** 

```latex
\documentclass{article}
\usepackage[showframe]{geometry}
\usepackage{pgfpages}

\def\pgfpagespoint#1#2{\pgfpoint{-#1}{#2}}

\pgfpagesuselayout{4 on 1}[a4paper,landscape]

\begin{document}

\pagedir TRT \bodydir TRT \pardir TRT 

\Huge 

Page 1
  
\newpage

Page 2

\newpage

Page 3

\newpage

Page 4

\end{document}

```

tex/latex/pgf/utilities/pgfpages.sty
```diff
@@ -34,6 +34,7 @@
 \pgfphysicalpageemptytrue
 \pgf@holdingphysicalpagefalse
 
+\let\pgfpagespoint\pgfpoint
```
muzimuzhi: With some auto detections:
```suggestion
\def\pgfpagespoint{%
  % Delay the definition of \pgfpagespoint in case the \pagedir is set at the
  % beginning of document, but still assume that the \pagedir is constant 
  % throughout the document.
  \pgfpages@test@pagedir\pgfpagespoint}

\def\pgfpages@test@pagedir{%
  \let\pgfpagespoint\pgfpoint
  \ifdefined\luatexversion
    % \pagedir returns a three-char string and \pagedirection returns an int
    \ifcase\pagedirection
        % 0, TLT (latin)
    \or % 1, TRT (arabic)
      \def\pgfpagespoint##1##2{\pgfpoint{##1}{##2}\pgf@x=-\pgf@x}%
    \or % 2, LTL (mongolian)
      % ... need any transformations?
    \or % 3, RTT (cjk)
      % ... need \pgf@x=-\pgf@x ?
    \fi
  \fi
}
```

muzimuzhi (commented):
Not sure if this is the way Henri will appreciate ... Nevertheless a way of auto detecting the page direction is suggested.

hmenke:
I don't think this will work properly when the page direction changes mid-document. Actually I'd like to see a refactor of `\pgfpageslogicalpageoptions` that takes multiple arguments for the different page directions.

seloumi:
Changes are for monolingual documents (latin or arabic), for others cases I don't know how it will be done.
Please add what you see suitable.

seloumi:
In most cases pagedir do not change in documents, and I think it should be mentioned that using `pgfpages` is limited to a single pagedir per document.

PR #1050, opened 2021-09-18
Tweaked `foreach` loops in example diagrams to avoid a floating-point rounding error.

3geek14:
…-point rounding error.

Signed-off-by: 3geek14 <nerd.of.pi@gmail.com>

An example in The Pic Syntax uses `{0, 0.1, ..., 1}` for a `foreach` loop. Due to floating-point arithmetic, this leaves out the value `1`. Rather than changing it to `1.01` or adding `, 1` to the list or using integers and dividing by 10 later…I changed the step to `0.25` to avoid the rounding error while maintaining the simplicity of the example.

**Checklist**

Please check the boxes below and [signoff your commits][git-s] to explicitly
state your agreement to the [Developer Certificate of Origin][DCO]:

- [ ✓ ] Code changes are licensed under [GPLv2][GPL] + [LPPLv1.3c][LPPL]
- [ ✓ ] Documentation changes are licensed under [FDLv1.2][FDL]

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-base-animations.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-base-animations.tex
@@ -339,7 +339,7 @@ key:
     animation list={0.333/\frac{1}{3},0.666/\frac{2}{3},1,1.333/1\frac{1}{3},1.666/1\frac{2}{3}},
 ]
 \tikz {
-  \foreach \i in {0,0.1,...,1} \draw (-0.9,.9-\i) -- ++(1.8,0);
+  \foreach \i in {0,0.125,...,1} \draw (-0.9,.9-\i) -- ++(1.8,0);
   \pgfanimateattribute{translate}{
     whom = node, begin on = {click},
     exit control={1}{0},
@@ -370,7 +370,7 @@ key:
     animation list={0.5,1,1.5,2,2.5},
 ]
 \tikz {
-  \foreach \i in {0,0.1,...,1} \draw (-0.9,.9-\i) -- ++(1.8,0);
+  \foreach \i in {0,0.125,...,1} \draw (-0.9,.9-\i) -- ++(1.8,0);
   \pgfanimateattribute{translate}{
     whom = node, begin on = {click},
     entry = {0s}{\pgfpointorigin},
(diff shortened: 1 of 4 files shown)


muzimuzhi:
Nice catch. I'm thinking if we should mention the rounding errors and their workarounds when `foreach` is first introduced in tutorial, more specific sec. 2.20 "Repeating Things: For-Loops", or just lead readers to sec. 88 "Repeating Things: The Foreach Statement", paragraph "The dots notation".

3geek14:
I'm not sure there's a good way to fit it into Karl's tutorial without it feeling out of place. I agree it's an important thing to remind users of, and I'm pretty certain I've made this exact error before. I'll try to spend more time thinking about this, and I'll make an issue or another PR if I come up with something I like.

3geek14:
There were a few other places where the same rounding error came up. I did a search for `0.1,` to find them all. (Thankfully, not too many coordinates have that, so it was reasonable to search.) I also changed to steps of 0.125 rather than 0.25, as it's visually more similar to the original, keeping a relatively dense step, while still (I think) easily interpreted by readers as a simple 1/n.

I also searched for `0.2,`, `0.3,`, and so on (excluding `0.5,`) to see if there were similar errors. The only other one I found is on the cover:

    \foreach \i in {0.5,0.6,...,2}
      \fill
        [white,opacity=\i/2,
         decoration=Koch snowflake,
         shift=(horizon),shift={(rand*11,rnd*7)},
         scale=\i,double copy shadow={
           opacity=0.2,shadow xshift=0pt,
           shadow yshift=3*\i pt,fill=white,draw=none}]
        decorate {
          decorate {
            decorate {
              (0,0)- ++(60:1) -- ++(-60:1) -- cycle
            } } };

For the cover, I think the visual design (for both picture and code) matters a lot more than it does in the examples I was editing, so I left it alone. I think my suggestion would be to use `{5,6,...,20}`, and then to replace `\i/2`, `\i`, and `3*\i` with `\i/20`, `\i/10`, and `0.3*\i`.

Of course, it's possible that it's intentional to have only 15 (rather than 16) snowflakes and for none of them to have opacity 1 (the highest currently around 0.95). In this case, I think it would be better to use `{0.5,0.6,...,1.95}` to make it look more intentional.

muzimuzhi:
At the beginning I thought maybe the two examples not related to animations could be adopted to show workarounds for rounding errors, but now I decided to keep drawing examples simple and perhaps extend the doc for `\foreach` to discuss more about the workaround, later or in your #1053.

Hence approved and I'm merging this.

PR #1053, opened 2021-09-20
Add warning about floating-point arithmetic to `foreach` introduction

3geek14:
Signed-off-by: 3geek14 <nerd.of.pi@gmail.com>

On #1050, @muzimuzhi suggested it might be worth adding a warning about floating-point errors in `foreach` loops to the tutorial where they are first introduced. I think my suggested change might be a bit cryptic about what sorts of rounding errors show up, but I don't know how explicit it's worth being here, when it's spelled out in far more detail in section 88, Repeating Things: The Foreach Statement. I also added to the example code a second loop that does have a rounding error preventing the final tick from being drawn.

**Checklist**

Please check the boxes below and [signoff your commits][git-s] to explicitly
state your agreement to the [Developer Certificate of Origin][DCO]:

- [ ✓ ] Code changes are licensed under [GPLv2][GPL] + [LPPLv1.3c][LPPL]
- [ ✓ ] Documentation changes are licensed under [FDLv1.2][FDL]

### DIFF
--- a/doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
@@ -1100,11 +1100,16 @@ however, be dimensionless real numbers) as in the following example:
 \end{codeexample}
 
 If you provide \emph{two} numbers before the |...|, the |\foreach| statement
-will use their difference for the stepping:
+will use their difference for the stepping, but be wary of floating-point
+rounding errors:
 %
 \begin{codeexample}[]
-\tikz \foreach \x in {-1,-0.5,...,1}
-       \draw (\x cm,-1pt) -- (\x cm,1pt);
+\begin{tikzpicture}
+  \foreach \x in {-1,-0.5,...,1}
+    \draw (\x cm,-1pt) -- (\x cm,1pt);
+  \foreach \x in {-1,-0.8,...,1}
+    \draw (\x cm,-7pt) -- (\x cm,-5pt);
+\end{tikzpicture}
 \end{codeexample}
 
 We can also nest loops to create interesting effects:
--- a/doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
+++ b/doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
@@ -1107,7 +1107,8 @@ rounding errors:
 \begin{tikzpicture}
   \foreach \x in {-1,-0.5,...,1}
     \draw (\x cm,-1pt) -- (\x cm,1pt);
-  \foreach \x in {-1,-0.8,...,1}
+  \foreach \x in {-1,-0.9,...,0}
+    % tick at 0 is not drawn due to rounding errors
     \draw (\x cm,-7pt) -- (\x cm,-5pt);
 \end{tikzpicture}
 \end{codeexample}


hmenke:
Could you please write a little more succinct commit messages? Commit messages can have more than one line, so something like this would be more appropriate:
```
manual: add warning about floating-point arithmetic #1050

Per a suggestion from @muzimuzhi, add a warning about floating-point arithmetic
to the first introduction of `foreach` in Karl's tutorial.
```
See also https://www.conventionalcommits.org/, although we do not entirely practice this here (yet).

3geek14:
Thanks for the link, I'll try to write better commit messages in the future (as well as try to remember to sign my commits). Since I'm not sure how to retroactively sign commits, my plan was to close this PR and submit a new one. Should I instead fix this one somehow?

hmenke:
You can run `git rebase --signoff -i origin/master` and replace all `pick` by `reword` to rewrite the commit messages and automatically signoff.

doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
```diff
@@ -1100,11 +1100,16 @@ \subsection{Repeating Things: For-Loops}
...
+rounding errors:
 %
 \begin{codeexample}[]
-\tikz \foreach \x in {-1,-0.5,...,1}
-       \draw (\x cm,-1pt) -- (\x cm,1pt);
+\begin{tikzpicture}
+  \foreach \x in {-1,-0.5,...,1}
+    \draw (\x cm,-1pt) -- (\x cm,1pt);
+  \foreach \x in {-1,-0.8,...,1}
+    \draw (\x cm,-7pt) -- (\x cm,-5pt);
```
muzimuzhi: Step `0.2` doesn't demonstrate the rounding error because it's rounded down:
```tex
 \foreach \x in {-1,-0.8,...,1} {\x, }
% gives
% 0, .2, 0.4, 0.59999, 0.79999, 0.99998,
```

I understand the consideration that `{-1,-0.9,...,1}` would draw too many ticks, so will `{-1,-0.9,...,0}` do the job?
```suggestion
  \foreach \x in {-1,-0.9,...,0}
    % tick at 0 is not drawn due to rounding errors
    \draw (\x cm,-7pt) -- (\x cm,-5pt);
```
image

muzimuzhi (changes requested):

doc/generic/pgf/text-en/pgfmanual-en-tutorial.tex
3geek14: Oops, thanks for the catch. I was initially nervous that a step of 0.1 would be too dense, rather than too many, but I think the added comment helps to draw attention to precisely where the missing tick should be, making the density less problematic. I also think that stopping at 0 rather than 1 makes the number of ticks more easily counted, in case people want to think "how many ticks should be drawn in total"?

muzimuzhi: Ah my poor English failed to properly express "too dense".

muzimuzhi:
Hmm applying suggested changes on github's webpage (by adding "Co-authored-by" trailer) will always _not_ add the "Signed-off-by" trailer, which makes the check failed.  Even worse when the two trailers are used together, whom the "I" used in [Developer Certificate of Origin](https://developercertificate.org/) is referring would be not unique / unclear.  Someone may interpret that only the committer is signed off, not including the one co-authored.

@hmenke What's your opinion in the long term?

hmenke:
Signed-off-by is not really important for small fixes. It's more important for new features. I don't want to get trolled again by someone who doesn't understand their own licensing terms.

hmenke:

> [!TIP]

All tests passed

PR #1094, opened 2021-12-09
Set up l3build infrastructure

josephwright:

This PR shows one approach to setting up `l3build`-based building and testing. It currently requires the
development version of `l3build`, with the intention of finalising discussion here before the latter is released

**Checklist**


josephwright:
Note I've not fiddled with the workflow files as the aim here is to agree on structures that work manually: automated testing will need to wait until that is done.

josephwright:
I've rebased onto the flattened `doc` tree. For the demo, I've used `./doc/generic/pgf` as the `docfiledir`, which means the PDF ends up next to the `.tex` file. I can switch that back if you like so that it ends up directly inside `./doc`: your call.

josephwright:
On the demo test files, I suspect you'll want a policy on naming. For LaTeX2e, we create a file `github-XXXX.lvt` for each bug/issue: these are focussed. For `l3draw`, I have one `.lvt` file per 'area', covering a submodule: those are pretty much identical to parts of the `pgf` layer. So perhaps something like `pgf-patterns.lvt`, etc., for the 'basic layer' and `tikz-XXX.lvt` for the 'interface' layer. (All just suggestions, of course.)

josephwright:
I've included the `color.cfg` file discussed in the previous PR, but I wonder if it's really needed if you are mainly building using a CI.

josephwright:
I've made a suggestion for the YAML setup, although I'm not 100% sure about how the push/tag split is set up here. I hope my suggested zip renaming is right: whilst one could add some extra code to `l3build` to pick up a tag, I've stuck to the basic approach that building a release just makes `pgf.<bits-and-pieces>.zip` and renamed it in the YAML.

.github/workflows/main.yml
```diff
@@ -188,7 +182,7 @@ jobs:
 
     - name: "Release: upload TDS zip"
       uses: actions/upload-release-asset@v1
-      if: matrix.engine == 'luatex' && startsWith(github.ref, 'refs/tags/')
+      if: tartsWith(github.ref, 'refs/tags/')
```
muzimuzhi:
```suggestion
      if: startsWith(github.ref, 'refs/tags/')
```

josephwright: Should be sorted

muzimuzhi: Ok now actions defined by `main.yml` are triggered again. I'm see why it fails.

.github/workflows/main.yml
```diff
@@ -100,35 +97,33 @@ jobs:
 
     - name: Generate the revision file
       run: |
-        texlua build.lua revisionfile
+        l3build revisionfile
```
muzimuzhi, line 100: CI fails at here. Local tests show `l3build revisionfile` returns exit code 0 but `texlua build.lua revisionfile` works fine. Hmm both generate the correct revision file.

Is a local function defined in `build.lua` automatically capable of a `<target>` in `l3build <target>`? Sorry I have very limited knowledge/experience about both Lua and l3build.

muzimuzhi: > Is a local function defined in `build.lua` automatically capable of a `<target>` in `l3build <target>`?

Seems `revisionfile` is already registered to `target_list`, see https://github.com/pgf-tikz/pgf/pull/1094/files#diff-9e218184846f8e9c379402b4d9ddea5abac901ca71d87dcaf61713467c19db42R96-R101

josephwright: I'd not added a return value, and it seems the YAML wants an explicit zero rather than 'N/A'.

muzimuzhi: > Local tests show `l3build revisionfile` returns exit code 0

typo: The exit code was 1, not 0.

Is it possible that it's `l3build` that sets the exit code 1? I notice [`main(target, names)`](https://github.com/latex3/l3build/blob/main/l3build-stdmain.lua#L154) in [`stdmain.lua`](https://github.com/latex3/l3build/blob/main/l3build-stdmain.lua) has this logic:
```lua
  if errorlevel ~= 0 then
    exit(1)
  else
    exit(0)
```
and since the default return value of a lua function is `nil` ...

josephwright:
I've not tried to split out workflows: if the aim is to make things run for less time, I'd put `l3build doc` and `l3build check` in separate jobs (probably using `l3build check -q -H`), and put `l3build ctan` into a workflow that's only triggered on tagging. That's probably easiest to see for the `latex3` repo (`latex2e` is rather tricky, while for simpler setups like `siunitx` I don't bother to save any time, and just use the `ctan` target.)

muzimuzhi:
Currently the manual is built twice in workflow `main.yml`, one in step "Build the manual", command `l3build doc`, another one in step "ZIP: Build", command `l3build ctan`. Is this the expected behavior? Because this almost doubles the whole running time.

Also now workflow `main.yml` fails at [step "CTAN: Validate"](https://github.com/pgf-tikz/pgf/runs/4484992299?check_suite_focus=true) with
```
curl: (26) Failed to open/read local data from file/application
Error: Process completed with exit code 1.
```
Probably caused by some simple mistakes but I have no idea.

josephwright:
@muzimuzhi The reason I've got `doc` and `ctan` is I was aiming to change as little as possible in the existing YAML: I can't say I follow it entirely!

On the return value: we will sort in `l3build`.

josephwright:
Right, I seem to have got the CI working :)

hmenke:
Great work! Thank you very much. I think this is almost good to go. The last thing on my agenda would be for you to rebase, squash and signoff. You can do it in three lines:
```
git pull origin master --rebase
git reset --soft origin/master
git commit -sm "refactor(ci): set up l3build infrastructure"
```

PR #1100, opened 2021-12-12
Adapt the repo to new l3build infrastructure

hmenke:

There is still some stuff to do:

- [x] Use Island of TeX Docker images for CI. This way we don't have to worry about TeX Live caching anymore. https://gitlab.com/islandoftex/images/texlive
- [x] Check feasibility of `l3build tag` instead of our custom `revisionfile`
- [x] Check feasibility of `l3build upload` instead of our custom GitHub Action
- [x] Abstractions over `lrbox`. Right now the test files are a bit too LaTeX specific.

I'm leaving these for later. It's more important that we can start writing tests now.

- [ ] Boilerplate for macro-based tests (e.g. for pgfmath and softpaths)
- [ ] Convert https://github.com/pgf-tikz/pgf/tree/master/source/generic/pgf/testsuite into actually usable l3build tests.

**Checklist**


muzimuzhi:
I'm also thinking of a set of per-format pre-processors so we can write tests for one format, e.g., `\pgfpicture ... \endpgfpicture` and run tests for all the supported formats.

josephwright:
@muzimuzhi If you want to use the same tests for multiple formats, I'd suggest doing something like the `d3...` ones for LaTeX3: have a short wrapper `.lvt` file for each setup, and a `.tex` file which contains the 'payloads'.

josephwright:
On the use of `lrbox`: this was a demo about one way to do catcode-dependent tests. As I showed, you could just `\showoutput` a suitable document. I used the LaTeX construct rather than a low-level one only because I've tried to stick to documented interfaces for the 'document level' syntax.

josephwright:
@hmenke I looked at doing a `pgf000` test (just loading), but I realise the current normalisation code doesn't like the `pgf` version strings. I'll look at sorting that, then will offer `pgf000` and `tikz000` (just checking loading). Beyond that, it would be handy to have a feeling for how you want tests dividing up. I'm sure I can manage a set of `\pgf...` ones based on the `l3draw` set.

.github/workflows/check.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Install PGF
+      run: |
+        tlmgr init-usertree --usertree $PWD
+        echo "TEXMFHOME=$PWD" >> $GITHUB_ENV
```
josephwright: I'm not sure this is actually needed: I didn't drop it in my PR, but that was mainly as I was changing as little as possible. Won't _hurt_, but it's also probably a distraction now.

.github/workflows/check.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    - name: Generate the revision file
+      run: |
+        export GIT_TAG=$(git describe --abbrev=0 --tags)
+        echo "GIT_TAG=$GIT_TAG" >> $GITHUB_ENV
+        l3build tag "$GIT_TAG"
+        cat tex/generic/pgf/pgf.revision.tex
```
josephwright, line 23: Not sure this is needed

.github/workflows/check.yml
```diff
@@ -0,0 +1,30 @@
...
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Install PGF
+      run: |
+        tlmgr init-usertree --usertree $PWD
+        echo "TEXMFHOME=$PWD" >> $GITHUB_ENV
+
+    - name: Generate the revision file
```
josephwright, line 16: I thing the test suite will run without this step, but I see why you want it in.

.github/workflows/check.yml
josephwright: (Same comment applies to other `.yml` files, of course.)

josephwright (approved):
I've made a few comments, but I think it looks fine. 

I might be tempted to rename the `.lvt` files to something like `pgf-example` and `tikz-example`.

.github/workflows/check.yml
hmenke: This is just for debugging to see whether the revision file has been generated correctly.

.github/workflows/check.yml
hmenke: Good point, I'll get rid of this.

build.lua
```diff
@@ -47,51 +47,47 @@ maxprintline = 9999
...
+  email = "pgf-tikz@tug.org",
+  license = "fdl;gpl2;lppl1.3c",
+  note = [[
+The release files are signed using a detached signature.  You can obtain the
+signature from the GitHub release page
 
-local git = {
-  tag = runcmd("git describe --abbrev=0 --tags"),
-  HEAD = runcmd("git rev-parse --abbrev-ref HEAD"),
+    https =//github.com/pgf-tikz/pgf/releases/download/${{ env.GIT_TAG }}/pgf_${{ env.GIT_TAG }}.ctan.flatdir.zip.sig]],
```
hmenke: @josephwright One last question: How do I get the `tagname` in here?

josephwright: At the moment there's not an easy way to do this that I can think of: will need some function you can call within the table.

hmenke: I saw there is `note_file` so I can just generate that during CI with the link from `upload-release-asset`.

doc/generic/pgf/CHANGELOG.md
```diff
@@ -0,0 +1,3294 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+Even though this release is not too heavy on user-facing additions it has seen a
+lot of contributed changes. Thanks to everyone who voluteered their time!
```
Mo-Gul:
```suggestion
lot of contributed changes. Thanks to everyone who volunteered their time!
```

doc/generic/pgf/CHANGELOG.md
```diff
@@ -0,0 +1,3294 @@
...
+### Added
+
+- l3build support for packaging PGF/TikZ
+- Add Matrix chat to README
+- Add rhombic polihedra #1022
+- Add Developer Certificate of Origin (DCO) to Pull Request template and enforce
+
+### Fixed
+
+- Wrap box content in `\pgfinterruptpicture` for shapes
```
muzimuzhi: This might be too vague, especially when in changelog.

```suggestion
- Wrap logic gate symbol in `\pgfinterruptpicture` for shapes in library
  `shapes.gates.logic.IEC`
```

doc/generic/pgf/CHANGELOG.md
```diff
@@ -0,0 +1,3294 @@
...
+- Remove superfluous `;` for shape `arrow box`
+- Remove superfluous `/utils/exec` in animations
+- Gobble `\pgf@stop` when parsing finishes in animations
+- Add missing `\pgf@sys@tonumber` before `<dimen>` in drivers and animations
+- Rewrite `dash expand off`
+- Better unknown key error msg in decorations
+- Fix `\let` in drivers for two csnames #1088
+- Protect against comma in pgfkeys arguments #389
+- Let active `"` expand to non-active `"` in pgfmath #1062
+- Eagerly evaluate `/tikz/rotate fit` #1071
```
muzimuzhi: It's noticeable that this also fixes the problem that `/tikz/rotate fit` avoids comma in its argument (a math expr). How about

```suggestion
- Protect against comma in `/tikz/rotate fit` argument and make it
  eagerly evaluated #1071
```

doc/generic/pgf/CHANGELOG.md
```diff
@@ -0,0 +1,3294 @@
...
+- Spelling and typo fixes in the manual
+- Update Debian installation instructions
+
+### Changed
+
+- Cleanup `\newif`s
+- Remove unused scripts
+- Remove experiments/ folder
+- Simplify loading by delegating to top-level files
+- Promote `Missing character` to errors
```
muzimuzhi:
```suggestion
- Promote `Missing character` to errors in building manual
```

muzimuzhi: Or add a `[docs]` tag?

PR #1132, opened 2022-01-30
fix: check for \relax in \pgfkeysifdefined

### LINKED ISSUE #1131: `pgfkeys`: Undefined `.@cmd` handler check is not robust
**Brief outline of the bug**

If the internal `.@cmd`-handler got defined to do nothing (or is `\relax`) a stray `\pgfeov` might end up in the input stream leading to a `Undefined control sequence` error.

```
\documentclass[]{article}

\usepackage{pgfkeys}

% just to make things trip
\csname pgfk@/foo/.@cmd\endcsname

\begin{document}
\pgfkeys{/foo}
\end{document}
```

---------------

Possible fix: Do `\let\pgfeov\relax` (I haven't thoroughly checked whether this might have side effects, but doubt it).

hmenke:

Fixes #1131

There are a lot of instances in pgfkeys where` \csname` is used, e.g. in `\pgfkeysgetvalueof`. This will `\let` the mentioned csname become `\relax` so we have to be prepared for this in `\pgfkeysifdefined`. Indeed, this was the original behavior before eTeX but it somehow got lost when transitioning to `\ifcsname`. The behavior is broken with eTeX since 641a8832.

TODO:

- [x] Write a unit test

**Checklist**


Skillmon (approved):

tex/generic/pgf/utilities/pgfkeys.code.tex
```diff
@@ -26,6 +26,13 @@
 \def\pgfkeys@empty{}
 \long\def\pgfkeys@firstoftwo#1#2{#1}
 \long\def\pgfkeys@secondoftwo#1#2{#2}
+\long\def\pgfkeys@ifcsname#1{%
+  \ifcsname#1\endcsname
+    \expandafter\pgfkeys@firstoftwo
+  \else
+    \expandafter\pgfkeys@secondoftwo
+  \fi
+}
```
muzimuzhi, line 35: The `\long` is unnecessary since `pgfkeys` is not ready to accept new paragraphs in key names.

tex/generic/pgf/utilities/pgfkeys.code.tex
```diff
@@ -180,7 +187,7 @@
 %
 % The length is \pgfkeysvalue{/tikz/length}.
 
-\def\pgfkeysvalueof#1{\csname pgfk@#1\endcsname}
+\long\def\pgfkeysvalueof#1{\csname\pgfkeys@ifcsname{pgfk@#1}{pgfk@#1}{relax}\endcsname}
```
muzimuzhi: and this `\long`

tex/generic/pgf/utilities/pgfkeys.code.tex
```diff
@@ -145,7 +152,7 @@
 %
 % \pgfkeyslet{/algo/swap}{\myswap}
 
-\def\pgfkeyslet#1#2{%
+\long\def\pgfkeyslet#1#2{%
   \expandafter\let\csname pgfk@#1\endcsname#2%
 }
```
muzimuzhi, line 157: To support `\pgfkeyslet{/foo}{\par}` and `\pgfkeysgetvalue{/foo}{\par}`, these two macros could have `\long`. (sorry this is a not-so-useful comment)

tex/generic/pgf/utilities/pgfkeys.code.tex
hmenke: I've had a discussion with Skillmon in the TeX.SX chat about this. It is in fact possible to smuggle `\detokenize{\par}` into the key name and although this is a rather academic corner case, it's better to make it `\long`.

Non-`\long` control sequences are actually becoming very rare. In expl3 `\cs_new:Npn` defaults to `\long`, `\newcommand` defaults to `\long` and in LuaMetaTeX all `\def`s are automatically `\long` with no way to opt out.

muzimuzhi: > Non-`\long` control sequences are actually becoming very rare. In expl3 `\cs_new:Npn` defaults to `\long`, `\newcommand` defaults to `\long` and in LuaMetaTeX all `\def`s are automatically `\long` with no way to opt out.

Reasonable. I will mark all the three conversations/comments resolved.

testfiles/gh-issue-1131.lvt
```diff
@@ -0,0 +1,22 @@
...
+% just to make things trip
+\csname pgfk@/foo/.@cmd\endcsname
+
+\begin{document}
+
+\START
+
+\BEGINTEST{pgfkeys: /.@cmd handler}
+\pgfkeys{/foo}
+\ENDTEST
```
muzimuzhi: I think I've (finally) got the point:
```tex
\pgfkeysgetvalue{/foo2}{\mycmd}
\pgfkeysifdefined{/foo2}{true}{false}
```
 - Before: true (`\pgfk@/foo2` is defined by the `\csname` used in `\pgfkeysgetvalue`)
 - After: false

Maybe this can be added to the test file, or even replace the current test 1 because of the invalid empty arg-spec.

testfiles/gh-issue-1131.lvt
```diff
@@ -0,0 +1,22 @@
...
+
+\BEGINTEST{pgfkeys: keys must be able to hold \string\relax}
```
muzimuzhi:
```suggestion
\BEGINTEST{pgfkeys: keys must be able to hold \relax}
```
`\BEGINTEST`'s first arg is `\detokenize`d before written to log.

testfiles/gh-issue-1131.tlg
```diff
@@ -0,0 +1,19 @@
...
+Type  H <return>  for immediate help.
+ ...                                              
+l. ...\pgfkeys{/foo}
+This error message was generated by an \errmessage
+command, so I can't give any explicit help.
+Pretend that you're Hercule Poirot: Examine all clues,
+and deduce the truth by order and method.
+============================================================
+============================================================
+TEST 2: pgfkeys: keys must be able to hold \string \relax
```
muzimuzhi:
```suggestion
TEST 2: pgfkeys: keys must be able to hold \relax 
```

tex/generic/pgf/utilities/pgfkeys.code.tex
```diff
@@ -379,7 +380,9 @@
 \def\pgfkeys@case@one{%
   \pgfkeysifdefined{\pgfkeyscurrentkey/.@cmd}%
   {\pgfkeysgetvalue{\pgfkeyscurrentkey/.@cmd}{\pgfkeys@code}%
-   \expandafter\pgfkeys@code\pgfkeyscurrentvalue\pgfeov}
+    \ifx\pgfkeys@code\relax\expandafter\pgfkeys@firstoftwo\else\expandafter\pgfkeys@secondoftwo\fi
+    {\pgfkeys@unknown}%
+    {\expandafter\pgfkeys@code\pgfkeyscurrentvalue\pgfeov}}
   {\pgfkeys@case@two}%
 }
```
muzimuzhi, line 397: With the change to `\pgfkeysifdefined` in https://github.com/pgf-tikz/pgf/pull/1132/commits/b8492921a47dba146c9bd1aaa2946082c1c37935, this newly added logic is unnecessary.

I know it's required to pass the current test 1, which is equivalent to
```tex
\pgfkeyslet{/foo/.@cmd}{\relax}
\pgfkeys{/foo}
```
but as I said in https://github.com/pgf-tikz/pgf/issues/1131#issuecomment-1025153232, in normal usage, there's no way for a `/.@cmd` subkey to be equal to `\relax` (as far as `\pgfkeysgetvalue` and the like are now `\csname` safe). One could wrongly write `\pgfkeyslet{/foo/.@cmd}{<token>}` in which the arg-spec of `<token>` doesn't end with `\pgfeov` and get similar errors.

This added logic `\ifx\pgfkeys@code\relax...` only save the case when `<token> == \relax`.

muzimuzhi (changes requested):

tex/generic/pgf/utilities/pgfkeys.code.tex
Skillmon: Since the aim was never to catch any wrong token, but just an erroneous let to `\relax` I think this could be fine. But you got a point with the situation now being that through official interfaces it shouldn't be possible to end up with such an accidental assignment. So I have no strong opinion on this (other than that there could be code outside the scope of "our" control that could do such assignments by unconditionally using `\csname`).

So the team should decide here: What is in the scope of these tests? Should the file be self-secure or also catch the most simple other error source. I guess everyone agrees on that the purpose shouldn't be detection of *every* error case of wrong assignment.

tex/generic/pgf/utilities/pgfkeys.code.tex
```diff
@@ -163,7 +170,7 @@
 %
 % \pgfkeysgetvalue{/tikz/swap}{\myswap}
 
-\def\pgfkeysgetvalue#1#2{\expandafter\let\expandafter#2\csname pgfk@#1\endcsname}
+\long\def\pgfkeysgetvalue#1#2{\expandafter\let\expandafter#2\csname\pgfkeys@ifcsname{pgfk@#1}{pgfk@#1}{relax}\endcsname}
```
muzimuzhi: Only a personal suggestion, maybe the code below is clearer? I have no idea which one is quicker.
```suggestion
\long\def\pgfkeysgetvalue#1#2{%
  \pgfkey@ifcsname{pgfk@#1}
    {\expandafter\let\expandafter#2\csname pgfk@#1\endcsname}
    {\let#2\relax}}}
```

tex/generic/pgf/utilities/pgfkeys.code.tex
hmenke: It makes for a much nicer error and an accidental `\relax` is easy to creep in. There are quite a few packages in TeX Live that access `\pgfk@...` directly.

tex/generic/pgf/utilities/pgfkeys.code.tex
muzimuzhi: similarly
```suggestion
\long\def\pgfkeysvalueof#1{%
  \pgfkeys@ifcsname{pgfk@#1}
    {\csname pgfk@#1\endcsname}{\relax}}
```
For compatibility concerns, leaving nothing seems to be less feasible than leaving `\relax`, when the checked key `#1` is undefined.

testfiles/gh-issue-1131.lvt
```diff
@@ -0,0 +1,26 @@
...
+\RequirePackage{pgfkeys}
+
+\begin{document}
+
+\START
+
+\BEGINTEST{pgfkeys: prevent csname from implicitly defining /foo}
+\pgfkeysgetvalue{/foo}{\mycmd}
+\pgfkeysifdefined{/foo}{\TYPE{true}}{\TYPE{false}}
+\ENDTEST
```
muzimuzhi, line 13: As another touched command, `\pgfkeysvalueof` worth to be tested as well:
```suggestion
\pgfkeysvalueof{/foo}
\pgfkeysifdefined{/foo}{\TYPE{true}}{\TYPE{false}}
\ENDTEST
```
or just `\TYPE{\pgfkeysifdefined{/foo}{true}{false}}`

muzimuzhi (commented):
Worth documenting the changed behavior of `\pgfkeysgetvalue` and `\pgfkeysvalueof`?

Sorry I didn't put all my thoughts in a single review.

tex/generic/pgf/utilities/pgfkeys.code.tex
hmenke: This is not feasible. It changes the number of expansion steps required to get to the result. A `\csname` being a primitive expands in a single step but if we put the check outside if would be `\pgfkeys@ifcsname` -> `\ifcsname` -> `\pgfkeys@firstoftwo` -> `\csname` which is suddenly four expansions. If clients rely on the number of `\expandafter`s necessary to get the value out this will be breaking for them.

I actually think I also have to partially revert my changes to `\pgfkeysifdefined`. It also requires one more step now.

For the non-expandable ones like `\pgfkeysgetvalue`, etc. this doesn't matter. We can turn the check inside out if that increases readability.

tex/generic/pgf/utilities/pgfkeys.code.tex
muzimuzhi: > There are quite a few packages in TeX Live that access `\pgfk@...` directly.

So this PR actually introduces a breaking change (that a not-so-right behavior since 2007 is now changed) which might affect more than I though before? I'd suggest to move the changelog entry under section `### BREAKING CHANGES`.

tex/generic/pgf/utilities/pgfkeys.code.tex
muzimuzhi: > It changes the number of expansion steps required to get to the result.

Ah I never thought from this aspect before (though it's possible to simulate by `\pgfkeys@expanded` but that will also complicate the implementation). Keeping compatibility for undocumented behavior is really a tough thing.

hmenke: `\pgfkeys@expanded` will not solve the problem easily becaues we don't want to expand what's stored in the key as well. :wink: These types of mistakes also happen to me quite frequently.

muzimuzhi: I had `\pgfkeys@unexpanded\expandafter{\csname pgfk@#1\endcsname}` in mind when commented but didn't typed it out. That's why I said "... will also complicate the implementation". Perhaps it's English that prohibits me from typing out everything I thought, or it's part of my bad habit. sorry

muzimuzhi:
@hmenke my bad, please revert my last force push

doc/generic/pgf/CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ lot of contributed changes. Thanks to everyone who volunteered their time!
   `\pgfrevision{,date}` and `\pgfversion{,date}` are synonyms for now, but the
   revision should eventually gain back its original meaning.  However, as of now
   this is not supported by l3build.
+- Many operations in `pgfkeys` used to use `\csname` directly which lets the
+  given csname become `\relax` in case it wasn't defined. This resulted in some
+  leakage into of accidentally `\relax`ed keys into the global scope. This has
+  been cleaned up a little. To preserve compatibility macros that used to expand
+  to a `\relax`ed csname now expand to a primitive `\relax`. However, this means
+  that the result of these macros must not be assigned to since this would
+  overwrite the primitve `\relax`.
```
muzimuzhi: It seems `\pgfkeysvalueof` and `\pgfkeysgetvalue` are the only two directly affected. Mention them here?

On line 21, is the first `into` after `leakage` redundant?

It takes me some time to understand the last two sentences, hence might take other readers longer. I understand a changelog entry should use only text descriptions whenever possible, but still ... . Maybe leave a comment containing an explanatory example under this PR and append the (unfortunately long) link to the changelog entry?

feel I've made too many comments (than acceptable) here 😞

tex/generic/pgf/utilities/pgfkeys.code.tex
Skillmon: Turning the test inside out for the non-expandable ones might be faster in the undefined case since TeX doesn't have to build `\relax` that way but instead in can be directly provided.

doc/generic/pgf/CHANGELOG.md
Skillmon: For the last sentence, maybe the following is easier to understand: "However, this means that you must not make assignments to the result of these macros' expansion since this would overwrite the primitive `\relax`".

Also for that only `\pgfkeysvalueof` is really affected, `\pgfkeysgetvalue` doesn't yield `\relax` but lets `\<user-provided-cs>` to `\relax` (huge difference) assigning to which is fine. And assigning to the result of `\pgfkeysvalueof` seems strange anyways (while I could imagine use cases I don't think those are common, feel free to correct me here).

However: I wonder whether that's the best behaviour anyway (I know, too late to change, but I'd prefer it threw an expandable error that the key isn't defined and return just nothing, and yes, that could be build in two steps of expansion).

muzimuzhi: > Also for that only `\pgfkeysvalueof` is really affected, [...]

@Skillmon You're right. Of course that includes both direct and indirect uses. I started to think the changed behavior and its subtle consequence can be added to docs for these macros and be mentioned in this changelog entry.

tex/generic/pgf/utilities/pgfkeys.code.tex
```diff
@@ -180,7 +190,8 @@
 %
 % The length is \pgfkeysvalue{/tikz/length}.
 
-\def\pgfkeysvalueof#1{\csname pgfk@#1\endcsname}
+\let\pgfkeys@relax\relax
+\long\def\pgfkeysvalueof#1{\csname\pgfkeys@ifcsname{pgfk@#1}{pgfk@#1}{pgfkeys@relax}\endcsname}
```
hmenke, line 194: I've added an alias for `\relax` for the unlikely event that someone actually does assign to this.

muzimuzhi (approved):

doc/generic/pgf/CHANGELOG.md
```diff
@@ -16,6 +16,18 @@ lot of contributed changes. Thanks to everyone who volunteered their time!
   `\pgfrevision{,date}` and `\pgfversion{,date}` are synonyms for now, but the
   revision should eventually gain back its original meaning.  However, as of now
   this is not supported by l3build.
+- Many operations in `pgfkeys` used to use `\csname` directly which lets the
+  given csname become `\relax` in case it wasn't defined. This resulted in some
+  leakage of accidentally `\relax`ed keys into the global scope. This has been
+  cleaned up a little. To preserve compatibility macros that used to expand to a
+  `\relax`ed csname now expand to a primitive `\relax`. This affects the
```
Skillmon, line 23: Outdated. Was changed to expand to `\pgfkeys@relax` (same four lines below, the last parts of this entry remain true however, this is a breaking change and one shouldn't make assignments to the results without testing for `\relax` or `\pgfkeys@relax`).

testfiles/gh-issue-1131.lvt
```diff
@@ -0,0 +1,26 @@
...
+
+\BEGINTEST{pgfkeys: keys must be able to hold \relax}
+\pgfkeyslet{/bar}{\relax} % or \pgfkeys{/bar/.initial=\relax}
```
Skillmon: Regarding the comment: `\pgfkeys{/bar/.initial=\relax}` doesn't `\let` to `\relax` but does `\def\<key-path>{\relax}`, so `<key-path>` doesn't have the `\meaning` `\relax`. More specifically: I'd expect `\pgfkeysifdefined` being false for `\pgfkeyslet{<path>}{\relax}` but true for `\pgfkeys{<path>/.initial=\relax}`. Maybe add a test for this.

doc/generic/pgf/CHANGELOG.md
Skillmon: Also: Should this get documented for `\pgfkeysvalueof` or should the documentation remain unchanged?

testfiles/gh-issue-1131.lvt
hmenke: I don't think letting a key to `\relax` should “undefine” it. For that it's better to let it to something that is actually undefined.

Skillmon: @hmenke agreed that `\pgfkeyslet{/bar}{\myundefinedmacro}` is better, but `\let\foo\relax` will have the same result for the majority of tests (and for the tests that have been implemented by now in `pgfkeys`).

hmenke: `\pgfkeysifdefined` has been using `\ifcsname` since 2007. I've not actually changed `\pgfkeysifdefined` but just cleaned up the accidental `\relax`ing. Now that I have squashed all the commits this should also be clearer.

PR #1144, opened 2022-03-09
Fix '.search also' with /handler config=full or existing

### LINKED ISSUE #1143: Handler config 'full or existing' can set non-existent keys when '.search also' is in effect.
**Brief outline of the bug**

With `/handler config=full or existing`, an assignment like `key/.style={stuff}` should only set the style if `key` exists in the current path or somewhere in the search paths set with `.search also`.  However, when one or more paths are passed to `.search also`, and `key` does not exist in the current path, `key/.style={stuff}` will always be executed it in the *first* search path, regardless of whether it exists there or not. It will never fail, like it should when the key on which the handler is called does not exist.

In the example below, the first case (with color=green) appears to work. The style `every foo` is set in `/main path`, because the style did not exist in `/second path`. However, in the second example you can see it fail: `every foo` is set in `/third path` although it did not exist there before, despite the handler config.

**Cause of the bug**

If you look at the log file, you can see the cause of the issue (thanks to the patched commands):
1. `.search also` executes `\pgfqkeys{/third path}{every foo/.style/.try/.expand once=value}` (defined in `\pgfkeys@searchalso@appendentry`).
2. `\pgfkeys@add@path@as@needed` prepends `/third path` and sets `\ifpgfkeysaddeddefaultpath=\iftrue` (in `\pgfkeys@addpath`).
3. `\pgfkeys@ifexecutehandler@handlefullorexisting` executes `.expand once` because it exempt from `/handler config`.
4. `.expand once` executes `\pgfkeysalso{/third path/every foo/.style/.try=value}`.
5. `\pgfkeysalso` calls `\pgfkeys@add@path@as@needed`, which sees an absolute path and sets `\ifpgfkeysaddeddefaultpath=\iffalse` (in `\pgfkeys@nevermind`).
6. `.try` calls `\pgfkeys@ifexecutehandler@handlefullorexisting` with `\pgfkeyscurrentpath=/third path/every foo/.style`, which succeeds because `\ifpgfkeysaddeddefaultpath` is false.

*Log file excerpt:*
```
[...]
pgfkeys@addpath: every foo/.style
pgfkeys@addpath: every foo/.style/.try/.expand once
pgfkeysalso: /third path/every foo/.style/.try={color=blue}
pgfkeys@nevermind: /third path/every foo/.style/.try
/third path/every foo/.style={color=blue}
pgfkeys@nevermind: /third path/every foo/.code
[...]
```

```latex
\documentclass{article}
\usepackage{pgfkeys}

\pgfkeys{
    /main path/foo/.code={\{\pgfkeysalso{/main path/every foo} foo\}},
    /main path/every foo/.style={color=red}, % It even works if this is commented out
    /main path/color/.code=color\{#1\},
    /handler config=full or existing,
}

\usepackage{etoolbox}
\makeatletter
\pretocmd\pgfkeys@nevermind{
    \immediate\write16{pgfkeys@nevermind: \detokenize{#1}}
}{}{}
\pretocmd\pgfkeys@addpath{
    \immediate\write16{pgfkeys@addpath: \detokenize{#1}}
}{}{}
\pretocmd\pgfkeysalso{
    \immediate\write16{pgfkeysalso: \expandafter\expandafter\expandafter\detokenize\expandafter\expandafter\expandafter{#1}}
}{}{}
\pgfkeys{/handlers/.style/.code=\immediate\write16{\pgfkeyscurrentpath/.style={#1}}\pgfkeys{\pgfkeyscurrentpath/.code=\pgfkeysalso{#1}}}
\makeatother

\begin{document}
\pgfkeys{
    /second path/.cd,
    /second path/.search also={/main path},
    every foo/.style={color=green}, % no 'every foo' in /second path
    foo
}
This worked, foo is green, not red

\pgfkeys{
    /second path/.cd,
    /second path/.search also={/third path, /main path},
    every foo/.style={color=blue},
    foo
}
This failed, foo is still green, not blue
\end{document}
```

fritzw:
Make .expand* keys use relative paths if the original path was relative.

Signed-off-by: Fritz Webering <fritz@webering.eu>

Fixes #1143

**Checklist**

### DIFF
--- a/doc/generic/pgf/CHANGELOG.md
+++ b/doc/generic/pgf/CHANGELOG.md
@@ -28,6 +28,10 @@ lot of contributed changes. Thanks to everyone who volunteered their time!
   case the value is undefined instead of a `\relax`ed csname. It has always been
   semantically wrong to assign to the result of `\pgfkeysgetvalueof`, but now it
   will have undesired side-effects. Therefore this change is breaking.
+- If `/handler config=full or existing` is active, `/.style={...}` assignments
+  can now fail if the style is not found in any `.search also` path. Previously,
+  the style was always set in the first search path, if it was not found in the
+  currently active path.
 
 ### Added
 
@@ -74,6 +78,7 @@ lot of contributed changes. Thanks to everyone who volunteered their time!
 - Form-only patterns have no specified color #1122 
 - Make `graphdrawing` work with `name prefix` and `name suffix` options #1087
 - pgfkeys was a bit too relaxed around `\relax` #1132
+- Fix `.search also` when `/handler config=full or existing` is active #1143
 
 ### Changed
 
@@ -90,6 +95,7 @@ lot of contributed changes. Thanks to everyone who volunteered their time!
 
 - 3geek14
 - BeneIII
+- Fritz Webering
 - graue70
 - Gábor Braun
 - Joel Coffman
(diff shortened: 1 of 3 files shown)


hmenke:

> [!TIP]

All tests passed

PR #1209, opened 2022-10-20
chore: add dependabot to keep actions up-to-date

muzimuzhi:

Some of the actions used by pgf is outdated and throwing warnings, see https://github.com/pgf-tikz/pgf/actions/runs/3242834076.
image

With an `dependabot.yml` config file and the dependabot enabled in [Insights](https://github.com/pgf-tikz/pgf/network/updates) (already enabled when I checked it), dependabot will open PR(s) to update outdated actions, periodically. Hope this would ease the effort to manually check and do necessary action updates.

The committed `dependabot.yml` is copied from section "Example `dependabot.yml` ..." in doc page referred below.

Reference:
- [Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) | GitHub Docs

**Checklist**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
--- a/doc/generic/pgf/CHANGELOG.md
+++ b/doc/generic/pgf/CHANGELOG.md
@@ -43,6 +43,7 @@ lot of contributed changes. Thanks to everyone who volunteered their time!
 - Add Developer Certificate of Origin (DCO) to Pull Request template and enforce
 - Add test set for `graphdrawing` (gd) 
 - pgfkeys gained support for loading libraries
+- Add dependabot to keep GitHub Actions up to date
 
 ### Fixed
 


muzimuzhi:
Sorry, it seems dependabot is a bit noisy. Should I config it to ignore patch and minor versions, like
```diff
diff --git a/.github/dependabot.yml b/.github/dependabot.yml
index df4d15b..c864d23 100644
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,6 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every week
       interval: "weekly"
+      ignore:
+        - dependency-name: "*"
+          update-types: ["version-update:semver-minor", "version-update:semver-patch"]
```

Also it's a pity that currently dependabot will always open a PR instead of an issue, which is softer. See feature request https://github.com/dependabot/dependabot-core/issues/2242.

hmenke:
I think it's fine as it is. We don't use a lot of actions and the one we do use don't update too often and are only pinned on major versions in the first place.

muzimuzhi:
> [...] the one we do use [...] are only pinned on major versions in the first place.

Ah yes. Action version will only be updated if a newer version with same type (or level, or [precision](https://github.com/dependabot/dependabot-core/blob/1d3a67f66ebc26bdd3fc628fbb8f512c9d2ec5ec/github_actions/lib/dependabot/github_actions/update_checker.rb#L115-L117) as how it's called in the source file) exists. A handy but un-documented feature.

PR #1212, opened 2022-10-25
IEC symbols: do not rotate content of ac source, dc source

knuesel:

This PR fixes the `ac source` and `dc source` circuit graphics to draw their content without rotation (see for example [here](https://www.google.com/search?q=ac+source+diagram&tbm=isch): the sine wave is always straight).

Currently these symbols work well when drawn on a horizontal wire:

```latex
\begin{tikzpicture}[circuit ee IEC]
  \draw (0,0) to [ac source={direction info={volt=3}}] (1,0)
              to [dc source={direction info={volt=3}}] (2,0);
\end{tikzpicture}
```

image

However, the common usage is to draw voltage sources on vertical wires, and in this case they look wrong:

```latex
\begin{tikzpicture}[circuit ee IEC]
  \draw (0,0) to [ac source={direction info={volt=3}}] (0,1)
              to [dc source={direction info={volt=3}}] (0,2);
\end{tikzpicture}
```

image

Adding a manual rotation can fix the content of the nodes, but messes with the annotations:

```latex
\begin{tikzpicture}[circuit ee IEC]
  \draw (0,0) to [ac source={point up,direction info={volt=3}}] (0,1)
              to [dc source={point up,direction info={volt=3}}] (0,2);
  % Trying with manual rotation of both symbol and annotation:
  \draw (2,0) to [ac source={point up,direction info={point up,volt=3}}] (2,1)
              to [dc source={point up,direction info={point up,volt=3}}] (2,2);
\end{tikzpicture}
```

image

A workaround was [proposed](https://tex.stackexchange.com/questions/28016/rotationally-correct-ac-source-symbol-in-tikz) on tex.stackexchange but it would be nice to have this fixed in TikZ, especially since these symbols are generally used on vertical wires.

With this PR it looks like this:

```latex
\begin{tikzpicture}[circuit ee IEC]
  \draw (0,0) to [ac source={direction info={volt=3}}] (0,1)
              to [dc source={direction info={volt=3}}] (0,2);
\end{tikzpicture}
```

image

**Checklist**

[x] Changelog updated
[x] Commit signed-off

### DIFF
--- a/tex/generic/pgf/frontendlayer/tikz/libraries/circuits/tikzlibrarycircuits.ee.IEC.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/libraries/circuits/tikzlibrarycircuits.ee.IEC.code.tex
@@ -416,10 +416,16 @@
     circuit symbol size=width 2 height 2,
     shape=generic circle IEC,
     /pgf/generic circle IEC/before background={
-      \pgfpathmoveto{\pgfqpoint{0.7pt}{0pt}}
-      \pgfpathquadraticcurveto{\pgfqpoint{ 0.35pt}{-0.5pt}}{\pgfqpoint{0pt}{0pt}}
-      \pgfpathquadraticcurveto{\pgfqpoint{-0.35pt}{ 0.5pt}}{\pgfqpoint{-0.7pt}{0pt}}
-      \pgfusepathqstroke
+      {
+        % First cancel rotation to have straight sine wave
+        \pgfgettransformentries{\a}{\b}{\c}{\d}{\x}{\y}
+        \pgfmathatantwo{\c}{\a}
+        \pgftransformrotate{\pgfmathresult}
+        \pgfpathmoveto{\pgfqpoint{0.7pt}{0pt}}
+        \pgfpathquadraticcurveto{\pgfqpoint{ 0.35pt}{-0.5pt}}{\pgfqpoint{0pt}{0pt}}
+        \pgfpathquadraticcurveto{\pgfqpoint{-0.35pt}{ 0.5pt}}{\pgfqpoint{-0.7pt}{0pt}}
+        \pgfusepathqstroke
+      }
     },
     transform shape,
   },
@@ -428,12 +434,18 @@
     circuit symbol size=width 2 height 2,
     shape=generic circle IEC,
     /pgf/generic circle IEC/before background={
-      \pgfpathmoveto{\pgfqpoint{ 0.7pt}{ 0.2pt}}
-      \pgfpathlineto{\pgfqpoint{-0.7pt}{ 0.2pt}}
-      \pgfusepathqstroke
-      \pgfpathmoveto{\pgfqpoint{ 0.7pt}{-0.2pt}}
-      \pgfpathlineto{\pgfqpoint{-0.7pt}{-0.2pt}}
-      \pgfusepathqstroke
+      {
+        % First cancel rotation to have straight equal sign
+        \pgfgettransformentries{\a}{\b}{\c}{\d}{\x}{\y}
+        \pgfmathatantwo{\c}{\a}
+        \pgftransformrotate{\pgfmathresult}
+        \pgfpathmoveto{\pgfqpoint{ 0.7pt}{ 0.2pt}}
+        \pgfpathlineto{\pgfqpoint{-0.7pt}{ 0.2pt}}
+        \pgfusepathqstroke
+        \pgfpathmoveto{\pgfqpoint{ 0.7pt}{-0.2pt}}
+        \pgfpathlineto{\pgfqpoint{-0.7pt}{-0.2pt}}
+        \pgfusepathqstroke
+      }
     },
     transform shape,
   },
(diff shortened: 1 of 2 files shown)


tex/generic/pgf/frontendlayer/tikz/libraries/circuits/tikzlibrarycircuits.ee.IEC.code.tex
```diff
@@ -416,10 +416,16 @@
...
     shape=generic circle IEC,
     /pgf/generic circle IEC/before background={
-      \pgfpathmoveto{\pgfqpoint{0.7pt}{0pt}}
-      \pgfpathquadraticcurveto{\pgfqpoint{ 0.35pt}{-0.5pt}}{\pgfqpoint{0pt}{0pt}}
-      \pgfpathquadraticcurveto{\pgfqpoint{-0.35pt}{ 0.5pt}}{\pgfqpoint{-0.7pt}{0pt}}
-      \pgfusepathqstroke
+      {
+        % First cancel rotation to have straight sine wave
+        \pgfgettransformentries{\a}{\b}{\c}{\d}{\x}{\y}
+        \pgfmathatantwo{\c}{\a}
```
hmenke, line 422: Might be better to think about how to disable `transform shape` for the `before background` path, rather than trying to counter an existing rotation. The proposed solution might result in chaotic behavior in rotated scopes. Have you tried that?

sijow: I'm not sure it's the best way to do it but it seems fairly safe to me: it removes the rotation component from the transformation matrix, so that the path is drawn without rotation, no matter how the scope is rotated. I haven't been able to find an example where it causes problem. For example it works fine in a scope rotated by 45 degrees:

image

On the other hand, commenting-out the `transform shape` in the current TikZ code creates problems with the annotations:

image

Here's the full code for these two tests:

<details><summary>figure.tex (click to expand)</summary>

```latex
\documentclass[crop,border=2mm,tikz]{standalone}
\usetikzlibrary{circuits.ee.IEC}

\tikzset{
  ac source graphic PR/.style={
    circuit symbol lines,
    circuit symbol size=width 2 height 2,
    shape=generic circle IEC,
    /pgf/generic circle IEC/before background={
      {
        % First cancel rotation to have straigh sine wave
        \pgfgettransformentries{\a}{\b}{\c}{\d}{\e}{\f}
        \pgfmathatantwo{\c}{\a}
        \pgftransformrotate{\pgfmathresult}
        \pgfpathmoveto{\pgfqpoint{0.7pt}{0pt}}
        \pgfpathquadraticcurveto{\pgfqpoint{ 0.35pt}{-0.5pt}}{\pgfqpoint{0pt}{0pt}}
        \pgfpathquadraticcurveto{\pgfqpoint{-0.35pt}{ 0.5pt}}{\pgfqpoint{-0.7pt}{0pt}}
        \pgfusepathqstroke
      }
    },
    transform shape,
  },
  ac source graphic notransform/.style={
    circuit symbol lines,
    circuit symbol size=width 2 height 2,
    shape=generic circle IEC,
    /pgf/generic circle IEC/before background={
      \pgfpathmoveto{\pgfqpoint{0.7pt}{0pt}}
      \pgfpathquadraticcurveto{\pgfqpoint{ 0.35pt}{-0.5pt}}{\pgfqpoint{0pt}{0pt}}
      \pgfpathquadraticcurveto{\pgfqpoint{-0.35pt}{ 0.5pt}}{\pgfqpoint{-0.7pt}{0pt}}
      \pgfusepathqstroke
    },
    %transform shape,
  },
}

\begin{document}

% Using code from this PR
\begin{tikzpicture}[circuit ee IEC,
                    set ac source graphic=ac source graphic PR]
  \draw (0,0) to [ac source={direction info={volt=3}}] (0,1);
  %
  \begin{scope}[xshift=3cm,rotate=45]
    \draw (0,0) to [ac source={direction info={volt=3}}] (0,1);
  \end{scope}
\end{tikzpicture}

% Using current TikZ code without "transform shape"
\begin{tikzpicture}[circuit ee IEC,
                    set ac source graphic=ac source graphic notransform]
  \draw (0,0) to [ac source={direction info={volt=3}}] (0,1);
  %
  \begin{scope}[xshift=3cm,rotate=45]
    \draw (0,0) to [ac source={direction info={volt=3}}] (0,1);
  \end{scope}
\end{tikzpicture}
\end{document}
```

</details>

sijow: Ah sorry I think I have misunderstood what you proposed. Here I tested removing the `transform shape` from the whole graphic instead of just the `before background` as you said.

If the `transform shape` does only a rotation then disabling it just for the path could work, but I have no idea how to do that... If it can also do for example a scaling, then I guess it should not be disabled (even just for the before background path).

sijow: Actually I don't think this implementation is appropriate: for some reason I wrote it assuming the transformation matrix is only a combination of rotation, scaling and translation, but it needs not be the case...

muzimuzhi:
I've reported (or recorded) the problem this PR tried to resolve as a bug report, in a perhaps more general way, see #1230.

Issue #1313, opened 2024-03-26
Error of foreach and node commands


psparkmath:
### Brief outline of the bug

I found that `foreach` command does not work well for `\node`.

I wrote the code `\draw (P1) foreach \n in {2,3}{ -- (P\i) };` to draw lines connecting nodes P1, P2, P3, consecutively. But, this code draw lines connecting P1 and P2 and connecting P1 and P3.

If you change `\node` part to `\coordinate`, then `foreach` command works well.

### Minimal working example (MWE)

```latex
\documentclass{article}
\usepackage{tikz}

\begin{document}
\begin{tikzpicture}
\foreach[count=\i] \n in {1,3,2}{
    \node (P\i) at (\i,\n) {P\i};
%   \coordinate (P\i) at (\i,\n);
%   \node at (P\i) {P\i};
}
%\draw (P1) -- (P2) -- (P3);
\draw (P1) foreach \i in {2,3}{ -- (P\i)}; % <--- HERE!
\end{tikzpicture}
\end{document}

jasperhabicht:
This is actually well-known. See, for example: https://tex.stackexchange.com/a/75152/47927

hmenke:
As documented.

![image](https://github.com/pgf-tikz/pgf/assets/1903556/9da35f97-9f64-4225-ac1d-d0532a7b64ab)

There is an undocumented option `foreach[scope iterations=false]` but it's undocumented for a reason, namely that in most of the cases it's broken.

muzimuzhi:
Would be fixed by #1304 too?

hmenke:
Yes. It's also a duplicate of https://github.com/pgf-tikz/pgf/issues/356 and https://github.com/pgf-tikz/pgf/issues/1303 and closely related to https://github.com/pgf-tikz/pgf/issues/1047.

PR #1304, opened 2024-01-08
Fix foreach not remembering nodes between iterations

loopspace:
**Motivation for this change**

Fixes #1303 by remembering the last node between iterations of a foreach loop on a path.

**Checklist**

Please [signoff your commits][git-s] to explicitly state your agreement to the [Developer Certificate of Origin][DCO]. If that is not possible you may check the boxes below instead:

- [X] Code changes are licensed under [GPLv2][GPL] + [LPPLv1.3c][LPPL]
- [X] Documentation changes are licensed under [FDLv1.2][FDL]

[git-s]: https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s
[DCO]: https://developercertificate.org
[GPL]: https://www.gnu.org/licenses/gpl-2.0.html
[LPPL]: https://www.latex-project.org/lppl/lppl-1-3c.txt
[FDL]: https://www.gnu.org/licenses/fdl-1.2.html

Issue #1303, opened 2024-01-08
\foreach behavior inside a \draw


alceu-frigeri:
### Brief outline of the bug

Hi,
I don't know if it qualifies as a bug, or a misunderstanding of mine about how  a ```foreach``` should work, when used inside a ```path```.
From my understanding, section 88 of the manual (pages 1001+), the foreach should just, mainly, expands the code inside the braces (if there is one, replacing the variables of the foreach).
In the example below, my understanding was that the 3 cases (commented out) should be equivalent. producing a line from (X) to node (nA)  node (nB) and finally coordinate (X).
case (1), the simplest, does just that, connecting the arcs (at node nA and node nB) as expected.
case (2), 'does the job', by means of three \draw commands 
  ```\draw (X) -- (nA) ;  \draw (nA) -- (nB); \draw (nB) -- (X)```

case (3) was the surprise for me, I expected it to be equivalent too (well, to be equal to case (1)) .. but I got a continuous line from (X) to (Y) ignoring the nodes (arcs) dimensions.

Is that the expected behavior? what am I missing on it? or is it a foreach miss behavior ?

For a better context: I'm using this construct with coordinates obtained from ```\draw[name intersections={of=...```
(intersections library). So I'm forced to use foreach statements to connect the many 'intersection' coordinates, in which I place an arc node in each intersection.

In my use case what I have implemented, for now, is a workaround based on case (2), better said, from the intersection coordinate/nodes list I generate a clist of node-pairs then I execute a final foreach with said list.
```latex
 \gdef\@tmpA{#3}
  \gdef\@ListA{}
  \foreach \x in {1,...,\csuse{#2T}} {\xappto{\@ListA}{\@tmpA/n#2-\x,}\xdef\@tmpA{n#2-\x}}
  \xappto{\@ListA}{n#2-\csuse{#2T}/#4}
  \foreach \xa/\xb in \@ListA {\draw[line cap=round] (\xa) -- (\xb);}
  ```
  (that's just to have an idea what I ended doing, ```#3```would be ```X```, ```#4``` would be ```Y```, ```#2T``` is the macro name with the number of intersections.. the ```n#2-\x``` are the intersection node names. It works, just fine, but it would be nicer not to have to use such a construct to just connect the nodes, and be able, for instance, to have something like
  ```latex
\draw (#3) \foreach \x in {1,...,\csuse{#2T}} { -- (\x) } -- (#4) ;
  ```

### Minimal working example (MWE)

```latex
[...]

Issue #1047, opened 2021-09-16
`\path ... foreach ...` doesn't work with option `turn`


muzimuzhi:
**Brief outline of the bug**

from TeX-SX question: https://tex.stackexchange.com/questions/615588/coordinate-turn-option-in-foreach

In short, value of `\tikz@tangent` is not smuggled outside the TeX-group added to each execution of foreach loop body.
https://github.com/pgf-tikz/pgf/blob/85d2c38b78407e5b1ba42140dfd11fcdf8df1edf/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex#L371-L384

Is saving/restoring `\tikz@tangant` in pgffor hooks, as what did for `\tikz@last[xy]` the recommended fixing approach?
https://github.com/pgf-tikz/pgf/blob/85d2c38b78407e5b1ba42140dfd11fcdf8df1edf/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex#L2582-L2616

**Minimal working example (MWE)**

```latex
\documentclass{article}
\usepackage{tikz}

\begin{document}
Expected:
\begin{tikzpicture}
  \draw[red] (0,0) -- (1,.2) -- ([turn]90:1cm) -- ([turn]90:1cm) -- ([turn]90:1cm) -- ([turn]90:1cm);
\end{tikzpicture}

Actual:
\begin{tikzpicture}
  \draw (0,0) -- (1,.2) foreach \i in {1,2,3,4} { -- ([turn]90:1cm) };
\end{tikzpicture}
\end{document}
```
![image](https://user-images.githubusercontent.com/6376638/133629291-84bf57c4-55fc-4f2d-a3f2-c78a1f663993.png)

hmenke:
Adding it to the hooks is not very elegant, but definitely much more elegant than just making `\tikz@tangent` global. Please go ahead.

muzimuzhi:
Or we could add sth like `\pgffor@smuggleone` or `\aftergroup\pgffor@smuggled` to `pgffor`, since I guess there will be more in tikz which need to be smuggled, e.g., `\tikz@timer` (not sure). It seems the new idea can avoid assigning things alternatively local and global.

rumout:
Hello,

Any update on this bug? I found myself caught on this bug today.

As a workaround, I placed the code above on the preamble without issues. However, I'm just loading tikz package without any others (other packages, libraries and such), so I'm concerned if this would raise some issue on other pgf/tikz parts...

[...]

PR #1316, opened 2024-04-12
Fix foreach in TikZ to remember more stuff between iterations

### LINKED ISSUE #356: Drawing paths that start at a node using \foreach
*Migrated from SourceForge*
Author: nzeh
Timestamp: 2015-05-14 14:29:46.942000

In the following code example, two curves from `(a)` to `(v2)` and from `(a)` to `(v3)` are created, but a single curve from `(a)` to `(v3)` through `(v2)` should be drawn.

    :::LaTeX
    \begin{tikzpicture}
      \coordinate (v1)  at (0,0);
      \coordinate (c21) at (1,0);
      \coordinate (c22) at (2,0);
      \coordinate (v2)  at (2,1);
      \coordinate (c31) at (2,2);
      \coordinate (c32) at (3,2);
      \coordinate (v3)  at (4,2);
      \node[draw,inner sep=0pt,minimum size=6pt,circle] (a) at (v1) {};
      \draw (a) \foreach \i in {2,3} {.. controls (c\i1) and (c\i2) .. (v\i)};
    \end{tikzpicture}

The correct path is drawn if the `\foreach` loop is manually expanded to

    :::LaTeX
    \draw (a) .. controls (c21) and (c22) .. (v2) .. controls (c31) and (c32) .. (v3);

The path is also drawn correctly using

    :::LaTeX
    \draw (v1) \foreach \i in {2,3} {.. controls (c\i1) and (c\i2) .. (v\i)};

but it is of course not shortened to start only at the node's boundary.  This suggests that `\foreach` gets confused when the starting point of a path is a node rather than a coordinate.

loopspace:

Fixes #356, #1047, #1303, #1313 by remembering more information between iterations of a foreach loop on a path, in particular the last node.  Also provides a more flexible interface to remember more macros & dimensions at both the code level and the user level (the latter via pgfkeys).

See https://github.com/pgf-tikz/pgf/pull/1304 for a previous version of this pull request where something went wrong with signing the commits.

**Checklist**

### DIFF
--- a/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
+++ b/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
@@ -2580,28 +2580,77 @@
 
 \def\tikz@fchar oreach{\tikz@foreach}%
 
+%
+% These are to allow \foreach to remember macros and dimensions between iterations
+%
+  
+\def\tikzforeach@smugglers@cove{}
+\let\tikztostart=\relax
+
+\def\tikzforeach@smuggle@macro#1{%
+  \pgfutil@ifx#1\relax{}{%
+    \expandafter
+    \pgfutil@g@addto@macro
+    \expandafter
+    \tikzforeach@smugglers@cove
+    \expandafter
+    {%
+      \expandafter\def\expandafter#1\expandafter{#1}%
+    }%
+  }%
+}
+
+\def\tikzforeach@smuggle@dimen#1{%
+  \pgfutil@ifx#1\relax{}{%
+    \expandafter
+    \pgfutil@g@addto@macro
+    \expandafter
+    \tikzforeach@smugglers@cove
+    \expandafter
+    {%
+      \expandafter#1\expandafter=\the#1\relax%
+    }%
+  }%
+}
+
+\def\tikzforeach@smugglers@loot{%
+  %
+  \tikzforeach@smuggle@macro\tikz@moveto@waiting%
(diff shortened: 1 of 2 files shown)


cfr42:
https://tex.stackexchange.com/q/760634/

hmenke:

> [!TIP]

All tests passed

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
```diff
@@ -2580,28 +2580,77 @@
...
+  %
+}
+
+\tikzset{
+  remember macro/.code={
+    \tikzforeach@smuggle@macro#1
+  },
+  remember dimension/.code={
+    \tikzforeach@smuggle@dimen#1
+  }
```
hmenke, line 2635: I think this is wrong. After the first `\gdef\tikzforeach@smugglers@cove{}` the entries remembered through these keys will be gone.

hmenke (changes requested):

tex/generic/pgf/frontendlayer/tikz/tikz.code.tex
loopspace: Before the `\gdef\tikzforeach@smugglers@cove{}` then the macro is expanded so all of the entries remembered are reinserted into memory.

https://github.com/loopspace/pgf/blob/33aff2d593d60382ba6f85c9753502fc4c206f4a/tex/generic/pgf/frontendlayer/tikz/tikz.code.tex#L2639

PR #1342, opened 2024-06-25
Fix broken example of simpledemo.lua to work with "simple demo layout"

hansonchar:

I tried out the example at *36.2.1 The Hello World of Graph Drawing* but it didn't compile:

```latex
\documentclass{standalone}
\usepackage{fp,pgf,tikz,xcolor}
\usetikzlibrary{graphs,graphdrawing} \usegdlibrary{simpledemo}
\begin{document}
\tikz [very simple demo layout]
  \graph { f -> c -> e -> a -> {b -> {c, d, f}, e -> b}};
\end{document}
```

It seems to me the culprit is the implementation in [SimpleDemo.lua#L25-L39](https://github.com/pgf-tikz/pgf/blob/master/tex/generic/pgf/graphdrawing/lua/pgf/gd/examples/SimpleDemo.lua#L25-L39) differs from that in the documentation. 

**Checklist**

### DIFF
--- a/doc/generic/pgf/pgfmanual-en-gd-algorithm-layer.tex
+++ b/doc/generic/pgf/pgfmanual-en-gd-algorithm-layer.tex
@@ -139,21 +139,6 @@ pgf.gd.interface.InterfaceToAlgorithms.declare {
   }
 }
 \end{codeexample}
-\directlua{
-pgf.gd.interface.InterfaceToAlgorithms.declare {
-  key = "very simple demo layout",
-  algorithm = {
-    run =
-      function (self)
-        local alpha = (2 * math.pi) / \luaescapestring{#}self.ugraph.vertices
-        for i,vertex in ipairs(self.ugraph.vertices) do
-          vertex.pos.x = math.cos(i * alpha) * 25
-          vertex.pos.y = math.sin(i * alpha) * 25
-        end
-      end
-  }
-}
-}
 
 This code \emph {declares} a new algorithm (|very simple demo layout|) and
 includes an implementation of the algorithm (through the |run| field of the
@@ -177,7 +162,7 @@ advertise this fact to the ``user''. In the case of \tikzname, this means that
 the option key |very simple demo layout| becomes available at this point and we
 can use it like this:
 %
-\begin{codeexample}[]
+\begin{codeexample}[preamble={\usetikzlibrary{graphs,graphdrawing} \usegdlibrary{simpledemo}}]
 \tikz [very simple demo layout]
   \graph { f -> c -> e -> a -> {b -> {c, d, f}, e -> b}};
 \end{codeexample}
@@ -185,7 +170,7 @@ can use it like this:
 It turns out, that our little algorithm is already more powerful than one might
 expect. Consider the following example:
 %
-\begin{codeexample}[]
+\begin{codeexample}[preamble={\usetikzlibrary{graphs,graphdrawing} \usegdlibrary{simpledemo}}]
 \tikz [very simple demo layout, componentwise]
   \graph {
     1 -> 2 ->[orient=right] 3 -> 1;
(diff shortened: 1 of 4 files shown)


hansonchar:
~~The generated manual in pdf can be downloaded at [here](https://github.com/hansonchar/pgf/actions/runs/9656370317).  I've also tested the changes locally and the example above now works.~~

I've subsequently forced push a better fix.  Waiting for Github workflow to complete.

hmenke:

> [!TIP]

All tests passed

doc/generic/pgf/pgfmanual-en-gd-algorithm-layer.tex
```diff
@@ -797,7 +782,7 @@ \subsection{Examples of Implementations of Graph Drawing Algorithms}
 \label{section-gd-examples}
 
 \includeluadocumentationof{pgf.gd.examples.library}
-\includeluadocumentationof{pgf.gd.examples.SimpleDemo}
+\includeluadocumentationof{pgf.gd.examples.simpledemo}
```
hmenke: I don't really get the point of this change. All the other layouts also follow CamelCase spellling.

tex/generic/pgf/graphdrawing/lua/pgf/gd/examples/simpledemo.lua
```diff
@@ -23,17 +23,14 @@ local declare = require "pgf.gd.interface.InterfaceToAlgorithms".declare
...
-          local radius = vertex.options['radius'] or g.options['radius']
-          vertex.pos.x = radius * math.cos(i * alpha)
-          vertex.pos.y = radius * math.sin(i * alpha)
+        local alpha = (2 * math.pi) / #self.ugraph.vertices
+        for i,vertex in ipairs(self.ugraph.vertices) do
+          vertex.pos.x = math.cos(i * alpha) * 25
+          vertex.pos.y = math.sin(i * alpha) * 25
         end
       end
   },
```
hmenke, line 36: I'd prefer that instead of changing `simple demo layout` into `very simple demo layout` you would add the `very simple demo layout` from `doc/generic/pgf/pgfmanual-en-gd-algorithm-layer.tex` alongside the unmodified `simple demo layout`. I'm not super familiar with the Lua graphdrawing stuff, but I think you can have multiple layouts in one file.

hmenke (changes requested):

tex/generic/pgf/graphdrawing/lua/pgf/gd/examples/simpledemo.lua
hansonchar: Before this PR, `SimpleDemo.lua` declared the key "simple demo layout". The manual examples, however, invoke "very simple demo layout" — a different key that was not declared anywhere prior to this PR. As a result, the manual examples failed to compile.

Simply adding "very simple demo layout" as a new, separate file alongside an unmodified `SimpleDemo.lua` would leave `SimpleDemo.lua` as dead code, just as unused as it is today. To actually fix this, either the manual's invocation needs to change to "simple demo layout" (matching what `SimpleDemo.lua` exports), or `SimpleDemo.lua` needs to export "very simple demo layout" instead. I took the latter approach.

doc/generic/pgf/pgfmanual-en-gd-algorithm-layer.tex
hansonchar: Renamed `simpledemo.lua` to `SimpleDemo.lua`. Prior to this PR, the manual referred to the file inconsistently — sometimes lowercase, sometimes camel case. Now all references are changed to use camel case.

tex/generic/pgf/graphdrawing/lua/pgf/gd/examples/simpledemo.lua
hansonchar: Actually what I said is not entirely true. Let me work on this a bit more. Please ignore the review request for now.

PR #1358, opened 2024-09-06
Add a new monotone interpolation plot handler

FrnchFrgg:
This handler converts each plot stream command into a curveto command, except for the first, which is converted to the previously specified action. The curveto control points are computed and adjusted so that the resulting curve respects the monotonicity of the points. If a point is a local extremum of the point set, it will also be a local extremum of the generated curve.

See https://en.wikipedia.org/wiki/Monotone_cubic_interpolation for an explanation of the algorithm. The last part uses \pgfpathcurveto instead of using Hermitte base functions.

The plot handler is also made available to datavisualization through the `smooth monotone line` visualizer option`


hmenke:

> [!TIP]

All tests passed

PR #1406, opened 2025-08-22
Circle suggestion

hammondkd:

This is a suggestion for the `circle` and `ellipse` shapes. In most other shapes I see in the manual (e.g., `trapezium`, `semicircle`, `isosceles triangle`, `kite`), the `base east`, `mid east`, `base west`, and `base east` anchors all end up on the shape boundary. In `circle` and `ellipse`, however, they are directly beneath `east` and `west`, outside the shape border. This patch would place them on the edge of the background path for circles and ellipses as well.

**Checklist**

### DIFF
--- a/doc/generic/pgf/CHANGELOG.md
+++ b/doc/generic/pgf/CHANGELOG.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Fix formatting of `default XXX, initially XXX` in key docs #1278
+- Moved `base west`, `base east`, `mid west`, and `mid east` anchors
+    onto the edge of the shape for `circle` and `ellipse` shapes
 
 ## [3.1.11] - 2025-08-14 Henri Menke
 
--- a/tex/generic/pgf/libraries/shapes/pgflibraryshapes.geometric.code.tex
+++ b/tex/generic/pgf/libraries/shapes/pgflibraryshapes.geometric.code.tex
@@ -91,18 +91,26 @@
   \anchor{mid west}
   {%
     \pgf@process{\radius}
-    \pgf@xa=\pgf@x%
+    \pgf@xa=\pgf@x
+    \pgf@ya=\pgf@y
     \pgf@process{\centerpoint}
-    \advance\pgf@x by-\pgf@xa%
-    \pgfmathsetlength\pgf@y{.5ex}
+    \pgf@xc=\pgf@x
+    \pgf@yc=\pgf@y
+    \pgfmathsetlength\pgf@y{0.5ex}%
+    \pgfmathsetlength\pgf@x{\pgf@xc
+        - \pgf@xa*cos(asin((\pgf@y-\pgf@yc)/\pgf@ya))}
   }%
   \anchor{base west}
   {%
     \pgf@process{\radius}
-    \pgf@xa=\pgf@x%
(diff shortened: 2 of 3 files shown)


hmenke:
The proposal is interesting, but the implementation is problematic.

1. Evaluating `cos(asin(x))` is very expensive. It's also equivalent to `sqrt(1-x^2)` which is maybe slightly less expensive(?). If you draw thousands of circles in a document (people do this) you'll absolutely notice the overhead.
2. `asin(x)` is only defined for -1 ≤ x ≤ 1. pgfmath is not super precise, so roundoff errors will cause failures.
3. The previous point is particularly important, because the argument of `asin` contains a division which is notoriously problematic in pgfmath and often the cause of `dimension too large`.

What do you think @muzimuzhi?

hammondkd:
@hmenke The extra evaluation time is only invoked if the user somehow accesses one of the anchors on every single circle. Simply drawing a circle will not trigger it. I created a grid of 2500 circles just to check, and it was actually faster with the `cos(asin(...` code in there (but I'm certain that's coincidence/not significant). When I explicitly accessed the `base east` anchor for all 2500 circles by putting a node at that anchor, it increased the time TeX took to generate a PDF from 8.98s to 9.52s. Your mileage may vary.

Good point about `sqrt(1-x^2) = cos(arcsin(x))`. I initially tried using `x = xc + sqrt(r^2 - (y-yc)^2)`, which is equivalent now that I think about it more, but that gave me a "dimension too large" error. In hindsight, I could have fixed that by normalizing `y-yc` first, but that introduces the same concern you had about division.

As for which is faster...I ran the same 2500 circle with `base east` access tests with `cos(asin((y-yc)/r))` and the same with `sqrt(1-sqrt(((y-yc)/r)^2))`, and which one is faster is a toss-up. They both have the same range issue, but that basically means that if the text's base is ever outside the circle, Bad Things Will Happen.

Anyway...just a thought.

hmenke:
Maybe we could use `cos(90*x)` as a poor approximation of `sqrt(1-x^2)`. That doesn't have problem as we go outside of `x ∈ [-1, 1]`. Or we straight-up go for `1-abs(x)`.

The approximation only needs to be good enough to not be visually discernible for circles between `1sp` and `\maxdimen`, right?

image

muzimuzhi:
I didn't have the time to dig into, only traversed all the shapes having `base east` anchor from the pgfmanual.

Other circle/ellipse-based shapes whose `base east` and the alike anchors are not on the shape boundary: `nand gate US`, `magnetic tape`, `circle split`, `circle solidus`, `ellipse split`, and `rounded rectangle`.

Meanwhile, these circle/ellipse-based shapes do have their `base east` and the alike anchors on the shape boundary: `semicircle` and `ellipse callout`.

hmenke:

> [!TIP]

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-shapes`

        </details>

    </details>

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-shapes`

        </details>

    </details>

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-shapes`

        </details>

    </details>

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-shapes`

        </details>

    </details>

      - <details open><summary><code>config-examples</code></summary>

          - `pgfmanual-en-library-shapes`

        </details>

    </details>

</details>

PR #1457, opened 2026-07-04
Extend the valid range of `\pgfmathscientific`

### LINKED ISSUE #1444: Normalization of number in scientific notation in pgfmath
This is probably not a bug, but it is quite a surprising thing (disclaimer: probably it is a case of "don't do that", but just to document it...

When using scientific notation, it would be nice to normalize the mantissa before the final conversion. The example below shows the problem:

```latex
\documentclass{article}
\usepackage{pgf}
\begin{document}
\pgfmathparse{500e-6}\pgfmathresult{} vs \pgfmathparse{5e-4}\pgfmathresult
\end{document}
```

Image

This is doubly surprising if you do that inside a `pgfplot` environment, because in that case the `addplot` works (I understand it is because it's using some fpu) and the `\pgfmathparse` not:

```latex
\documentclass[border=10pt]{standalone}
\usepackage{pgfplots}\pgfplotsset{compat=newest}
\begin{document}
    %% use 0.0005 or 0.5e.3 and it works
    \newcommand{\valueB}{500e-6}%
    % \newcommand{\valueB}{0.5e-3}%
    \begin{tikzpicture}[declare function={
        NL(\t)=4 + 0.03*(\t) - \valueB*((\t-50)^2);}
    ]
    \begin{axis}[ xmin=0, xmax=90, domain=0:80, ]
        \addplot[color=red, ultra thick] {NL(x)}; % works always
        \pgfmathsetmacro{\Value}{NL(20)} % fails with \valueB 500e-6
        \node [draw,pin=above:\Value] at (20, \Value) {};
    \end{axis}
    \end{tikzpicture}
\end{document}
```

Image

hmenke:

Fixes #1444

We could potentially gain an even wider range by pre-normalizing the mantissa against the exponent, but the new code is already complex enough.

The diff is best viewed with whitespace changes hidden: https://github.com/pgf-tikz/pgf/pull/1457/changes?w=1

**Checklist**


Rmano (approved):
Not sure if I understand the low level TeX code fully, but I've not seen anything strange.

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
```diff
@@ -135,22 +135,25 @@
...
         \ifpgfmath@in@%
         \else%
-            \edef\pgfmath@temp{\pgfmath@temp.0}%
+            \edef\pgfmath@sci@mantissa{\pgfmath@sci@mantissa.0}%
         \fi%
-        \expandafter\pgfmath@scientific@@\pgfmath@temp00000000000\pgfmath@}
+        \expandafter\pgfmath@scientific@@\pgfmath@sci@mantissa00000000000\pgfmath@}
+
+\def\pgfmath@scientific@sign@mantissa#1#2\pgfmath@{%
+    \ifx#1-
```
Skillmon: There's a trailing space here

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
```diff
@@ -135,22 +135,25 @@
...
-        \pgfmath@x#1pt\relax%
-        \ifdim\pgfmath@x<0pt\relax%
-            \pgfmath@x-\pgfmath@x%
-            \edef\pgfmath@sci@sign{-}%
-            \expandafter\edef\expandafter\pgfmath@temp\expandafter{\expandafter\pgfmath@gobbleone#1}%
-        \else%
-            \edef\pgfmath@sci@sign{+}%
-            \edef\pgfmath@temp{#1}%
-        \fi%
+        \expandafter\pgfmath@scientific@sign@mantissa\pgfmath@sci@mantissa\pgfmath@
```
Skillmon, line 140: This gives a low level TeX error on every `#1` expanding to a blank or empty list (which would be a wrong input so this might be fine).

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
```diff
@@ -135,22 +135,25 @@
...
         \expandafter\pgfmath@in@\expandafter.\expandafter{#1}%
         \ifpgfmath@in@%
         \else%
-            \edef\pgfmath@temp{\pgfmath@temp.0}%
+            \edef\pgfmath@sci@mantissa{\pgfmath@sci@mantissa.0}%
         \fi%
-        \expandafter\pgfmath@scientific@@\pgfmath@temp00000000000\pgfmath@}
+        \expandafter\pgfmath@scientific@@\pgfmath@sci@mantissa00000000000\pgfmath@}
+
+\def\pgfmath@scientific@sign@mantissa#1#2\pgfmath@{%
```
Skillmon: This macro in its current form depends on the fact that there is at most a single leading `-` for negative numbers and no sign for positive numbers, I don't know enough about the inner workings of `\pgfmathparse` to know whether that assumption always holds.

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
```diff
@@ -160,41 +163,84 @@
...
-        \edef\pgfmath@sci@factor{\the\c@pgfmath@countc}%
-        \ifnum\c@pgfmath@countb<0\relax%
-            % xE-y: easy...
-            \pgfmath@x\pgfmath@sci@int.\pgfmath@sci@mantissaA pt\relax%
-            \divide\pgfmath@x\pgfmath@sci@factor\relax%
+        \ifnum\c@pgfmath@counta<-9
+            % 10^10 exceeds the range of a TeX count, and any value that
+            % fits a \dimen rounds below 1sp once divided by it, so flush
+            % such tiny numbers to zero.
+            \pgfmath@x0pt
```
Skillmon, line 183: Is the input guaranteed to be the expanded value of a `\dimen`? If not and the input can indeed be more or less arbitrary then `(\number\maxdimen) * 1e-10` is roughly 0.1.

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
```diff
@@ -160,41 +163,84 @@
...
+            % Calculating 10^|exponent| using \pgfmathpow@ would overflow already for
+            % exponents of 5 or larger, because that exceeds \maxdimen.  We can work around
+            % that to some extent by calculating 10^|exponent| by repeated multiplication of
+            % 10 in a loop with a TeX count that has a larger range than a \dimen and allows
+            % us to go all the way to 10^10.
+            \c@pgfmath@countc1
+            \pgfutil@loop\ifnum\c@pgfmath@counta>0
+                \multiply\c@pgfmath@countc10
+                \advance\c@pgfmath@counta-1
+            \pgfutil@repeat
```
Skillmon: Not sure whether you want to do that but this here is a bit faster starting at exponents ≥2 (benchmarked in pdfTeX) and also doesn't need any auxiliary functions:

```suggestion
            \begingroup
                \aftergroup\c@pgfmath@countc\aftergroup1%
                \pgfutil@loop\ifnum\c@pgfmath@counta>0
                    \aftergroup0%
                    \advance\c@pgfmath@counta-1
                \pgfutil@repeat
            \endgroup
            \relax
```

If you had an expandable loop instead things would be even faster than this.

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
```diff
@@ -160,41 +163,84 @@
...
+                \advance\c@pgfmath@counta\c@pgfmath@countc\relax
+                \divide\c@pgfmath@counta\pgfmath@sci@factor\relax
+                \pgfmath@x\pgfmath@sci@sign\the\c@pgfmath@counta sp
+            \else%
+                % xE+y: in contrast to \dimen calculations that signal "Dimension
+                % too large" the count multiplications below wrap silently on
+                % 32-bit overflow, which for an out-of-range value could yield a
+                % wrong, in-range-looking result.  So we first detect overflow by
+                % checking that the mantissa does not exceed \maxdimen/factor (in
+                % sp) and raise an error before performing the calculation.
```
Skillmon, line 222: I haven't yet mentally stepped through the processing following this comment, but I think your borders are more strict than necessary. I'll tackle this later, but wanted to leave the comments up to this point for you to see :)

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
hmenke: Good point. The way it's called right now by `\pgfmathparse` is

https://github.com/pgf-tikz/pgf/blob/dc56af18b72255bede59800928f9a2bab1fe35e0/tex/generic/pgf/math/pgfmathparser.code.tex#L598-L609

where `\pgfmath@number` is just a macro so can basically contain anything. However, sometimes it's also used like `\pgfmath@dimen=\pgfmath@number pt`, so it should probably always hold a valid TeX number.

Could you suggest a good alternative?

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
hmenke: Yes, I know this could in principle be done, but it would require normalizing the mantissa against the exponent first, making the code even more complicated than it already is. Do you think there is a need for an even wider input range?

Skillmon (approved):

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
Skillmon: Given that this is a big improvement already it's probably not worth the hassle. But I'd leave a comment explaining the reasoning and that this is a known edge case that could in principle be handled.

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
Skillmon: You'd need to step through the head of the argument counting `-` signs, dropping `+` signs, and then evaluating `\ifodd` on your temporary counter.

Something like this:

```latex
\def\pgfmath@scientific@sign@mantissa#1\pgfmath@{%
  \pgfmath@counta=0
  \pgfmath@scientific@sign@mantissa@#1\pgfmath@
}
\def\pgfmath@scientific@sign@mantissa@#1{%
  \ifx\pgfmath@#1%
    \pgfmath@error{Missing number in input}% input was only signs or spaces
  \fi
  \ifx-#1%
    \advance\pgfmath@counta1
    \expandafter\pgfmath@gobbleone
  \else\ifx+#1%
    \expandafter\expandafter\expandafter\pgfmath@gobbleone
  \else
    \expandafter\expandafter\expandafter\pgfmath@scientific@sign@mantissa@end
  \fi\fi
  #1\pgfmath@scientific@sign@mantissa@
}
\def\pgfmath@scientific@sign@mantissa@end#1\pgfmath@scientific@sign@mantissa@#2\pgfmath@{%
  \edef\pgfmath@sci@sign{\ifodd\pgfmath@counta-\else+\fi}%
  \edef\pgfmath@sci@mantissa{#1#2}%
}
```
Decide for yourself whether that's worth the hassle.

tex/generic/pgf/math/pgfmathfunctions.misc.code.tex
hmenke: Do you still want to have a go at this?

Skillmon: Oh sorry, forgot to resolve this convo! I did that two weeks ago when I approved :)

PR #1459, opened 2026-07-06
Set of small fixes

### LINKED ISSUE #437: [manual] The order of transformations in the description of the \pgfshadepath algorithm
*Migrated from SourceForge*
Author: evan-aad
Timestamp: 2017-07-27 07:09:23.359000

The shading algorithm given in the description of the \pgfshadepath command in subsection 109.3 ('Using Shadings') of chapter 109 ('Shadings') of the TikZ & PGF manual for version 3.0.1a reads (pp. 1085-1086):

> ... the center of the shading is translated (moved) such that it lies on the center of the bounding box of the path. The low-level coordinate system is also scaled such that the shading “covers” the path (the details are a bit more complex, see below). Then, the coordinate system is rotated by 〈angle〉.

Thus, according to the manual the order of transformations is: 1. shift, 2. scale, 3. rotate. However, in the source code (tex/generic/pgf/basiclayer/pgfcoreshade.code.tex) a different order is specified in the comment attached to the implementation of the \pgfshadepath command:

> The shading will be rotated by #2 and then rescaled so that it completely covers the path. 

Thus, according to the documentation inside the source code, the order of transfromations is: 1. rotate, 2. scale.

Since matrices in general, and transformation matrices in particular, are generally not commutative, the order of application matters!

A perusal of the source code reveals that the system-layer transformations are applied in the order 1. shift, 2. scale, 3. rotate. However since system-layer transformations affect the coordinate-system rather than the graphical objects, the  effect on the coordinates that his sequence have is the opposite, so that a user who is not familiar with system-layer transformations (and someone reading chapter 109 should not be expected to be familiar with these details), will perceive the effects of this algorithm as though the order of transformations were as described in the source-code comment, namely: 1. rotation, 2. scale, 3. shift, and will therefore be confused by the algorithm described in the manual.

In my opinion, the manual should be corrected to state the order of transfromations as 1. rotate, 2. scale, 3. shift. I personally was very confused when I read this algorithm in the manual and tried to compare it to the actual results of shading a path, and it wasn't until I delved into the source-code and read about PDF canvas transfromations that I realized what was really going on.

### LINKED ISSUE #611: docs of z axis default vector
*Migrated from SourceForge*
Author: kryde
Timestamp: 2014-09-06 07:00:48.123000

It'd be good if the manual said a little more about the nature of the default z axis vector.  

I thought to use some "rotate around" etc to draw in 3D, and then set a viewing angle with further overall "rotate around"s.  If I'm not mistaken the default z is like a shear, whereas when rotating coordinates a z=0,0 for perpendicular to the page would be wanted.  Is that right?  The manual could give some guidance if so or if not.

As a concrete suggestion perhaps something like below for section "25.2 The XY- and XYZ-Coordinate Systems" item "/tikz/z".  Or a cross reference if this already exists but I did not find it:

The default z=(3.85mm,-3.85mm) behaves as a shear to make the z part of coordinates visible when projected onto the flat page.

z=(0,0) gives a z axis pointing directly out of the page.  This is good if you're applying your own rotations to construct and then view a 3-D object from some angle.  (View in the sense of project the rotated parts onto the page as the XY plane.)

### LINKED ISSUE #1142: \x not used in in Branching Statements example
Version: `3.1.9a`

`\x` is defined but never used in 58.5 Branching Statements example (see [comment](https://github.com/pgf-tikz/pgf/commit/d2d2f3d19768e2034aa20fbdaca39a570e2374bd#r66902066)).

https://github.com/pgf-tikz/pgf/blob/b79c4c5c632110079c3827ce4963a27e6d5d1742/doc/generic/pgf/pgfmanual-en-library-math.tex#L364-L382

### LINKED ISSUE #1307: Adjustments for index
### Version

3.1.10

### Details

This is rather a suggestion for the manual: Add bookmarks for letter groups of the index to for easier navigation.

I've played with the code from the repro and it seems to work just fine. As my git foo is poor, I'll currently not be doing a PR, but will describe what I did.

# build.lua

Add the line

    indexstyle = "pgfmanual-index.ist"

And change `docfiles` to
```
docfiles =
  {
    "RELEASE_NOTES.md", "description.html", -- Part of the release script
    "color.cfg", "pgfmanual.cfg", "images", "plots", "pgfmanual-index.ist", "*.tex" -- Build the manual
  }
```
(added `pgfmanual-index.ist`)

# pgf/doc/generic/pgf/pgfmanual-en-main-body.tex

Replace the line

    \usepackage{imakeidx}

by

    \usepackage{makeidx}

(you're building using `l3build` which runs `makeindex` so there is no need for `imakeidx`)

# Add .ist file

In the same folder as `pgfmanual.tex` add the file `pgfmanual-index.idx` with the contents

```
preamble "\\begin{theindex}\n\\providecommand{\\doidxbookmark}[1]{\\pdfbookmark[2]{#1}{#1-idx}}\n"
headings_flag 1
heading_prefix "\\doidxbookmark{"
heading_suffix "}"
```

Then compile the manual using 

    l3build doc

By default `l3build` uses `gind.ist` which has a different look than the `makeindex` default (which is what the `imakeidx` uses), the style above just adjusts the makeindex defaults a bit in order to add bookmarks to each letter group, but will not print them (feel free to adjust this) so the printed index will look just like it does in the current release.

### LINKED ISSUE #1327: Suggestion: add mention to \pgfinterruptpath in \pgfextra's documentation
As in the title. If I understood correctly that's how it should be used.

https://tikz.dev/base-scopes#pgf.pgfinterruptpath

Side note: the answer in [Adding Tikz node in pgfextra triggers double drawing of containing path - TeX - LaTeX Stack Exchange](https://tex.stackexchange.com/questions/131077/adding-tikz-node-in-pgfextra-triggers-double-drawing-of-containing-path) uses `\pgfinterruptpath` without the environment form. I'm not sure if that is supposed to be supported.

### LINKED ISSUE #1384: pgfutil-plain.def has dulicate def'n of \pgfmath@tempdima
### Brief outline of the bug

I'm using texlive on Ubuntu.  The distribution id is `2024.20240706`.  In the file `pgfutil-plain.def`, near line 230, I happened to see the following code.   I'm guessing the second line is an unintentional duplicate.   Maybe intention was to define `dimb` ?

```
\let\pgfmath@tempdima\pgfutil@tempdima
\let\pgfmath@tempdima\pgfutil@tempdima
\let\pgfmath@tempcnta\pgfutil@tempcnta
\let\pgfmath@tempcntb\pgfutil@tempcntb
```

### Minimal working example (MWE)

N/A

### LINKED ISSUE #1391: Math active colon in node text causes infinite loop
### Brief outline of the bug

Originally reported to https://github.com/latex3/mathtools/issues/67.

### Minimal working example (MWE)

```tex
\documentclass{article}

\usepackage{mathtools}
\usepackage{tikz}

\mathtoolsset{centercolon}

\begin{document}

\begin{tikzpicture}
\tracingall
  \node {$f:A$};
\tracingnone
\end{tikzpicture}

\end{document}
```

The infinite loop happens at `:` used in node text, which is set to math active by `\mathtoolsset{centercolon}`.
```tex
~.:->\tikz@nonactivecolon

~..\tikz@nonactivecolon ->:
{the character :}

~..:->\tikz@nonactivecolon

~..\tikz@nonactivecolon ->:
{the character :}
```

The tikz path parser tried to work with both normal and active colons (and several other punctuations), but looks like it doesn't work with math active colon, used in node text yet.

This had been reported to TeX-SX years ago https://tex.stackexchange.com/questions/427611/pdflatex-hangs-on-centercolon-plus-tikz.

hmenke:

Fixes #611  
Fixes #437  
Fixes #1142  
Fixes #1307  
Fixes #1327  
Fixes #1384  
Fixes #1391  

**Checklist**

### DIFF
--- a/build.lua
+++ b/build.lua
@@ -12,7 +12,7 @@ docfiledir = "./doc/generic/pgf"
 docfiles =
   {
     "RELEASE_NOTES.md", "description.html", -- Part of the release script
-    "color.cfg", "pgfmanual.cfg", "images", "plots", "*.tex" -- Build the manual
+    "color.cfg", "pgfmanual.cfg", "pgfmanual.ist", "images", "plots", "*.tex" -- Build the manual
   }
 tdsroot = "generic"
 typesetfiles = {"pgfmanual.tex"}
--- a/doc/generic/pgf/CHANGELOG.md
+++ b/doc/generic/pgf/CHANGELOG.md
@@ -36,6 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add a new monotonic interpolation plot handler #1358
 - The keys `actualtext`, `alt`, and `artifact` are now pre-defined
   in both `tikz` and `pgf` as  no-ops for usage in tagging #1370 #1453
+- Documentation of the default z-axis vector #611
 
 ### Changed
 
--- a/doc/generic/pgf/CHANGELOG.md
+++ b/doc/generic/pgf/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   apply a single definition of options to multiple examples.
 - Support directions in LuaTeX and LuaMetaTeX
 - Fix unused variable in tikzmath example #1142
+- Correct source comment on `\pgfshadepath` #437
 
 ### Added
 
--- a/doc/generic/pgf/CHANGELOG.md
+++ b/doc/generic/pgf/CHANGELOG.md
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support directions in LuaTeX and LuaMetaTeX
 - Fix unused variable in tikzmath example #1142
 - Correct source comment on `\pgfshadepath` #437
+- Mention `\pgfinterruptpath` in the documentation of `\pgfextra` #1327
 
 ### Added
 
--- a/doc/generic/pgf/CHANGELOG.md
+++ b/doc/generic/pgf/CHANGELOG.md
@@ -39,6 +39,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The keys `actualtext`, `alt`, and `artifact` are now pre-defined
   in both `tikz` and `pgf` as  no-ops for usage in tagging #1370 #1453
 - Documentation of the default z-axis vector #611
+- The manual now has PDF bookmarks for the index sections #1307
 
 ### Changed
 
--- a/doc/generic/pgf/CHANGELOG.md
+++ b/doc/generic/pgf/CHANGELOG.md
@@ -49,6 +49,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - The long deprecated `vtex` and `textures` drivers have been removed
 - The long deprecated `pgfutil-common-lists` (part of pgfplots since 1.18.2) has been removed
+- Removed duplicate definition of `\pgfmath@tempdima` in pgfutil #1384
 
 ### Contributors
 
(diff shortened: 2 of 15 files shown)


Mo-Gul (approved):
As you know I only understand the "doc" parts, but they are fine.

muzimuzhi (approved):

PR #1467, opened 2026-08-03
docs: use "new" `circle` syntax, part 2

muzimuzhi:
**Motivation for this change**

Follow-up to PR #1460, so also part of #999.

- These cases are found with `rg --multiline '\b(circle|ellipse)\s*\(' -g '!**/*.tlg'`, then checked one-by-one manually.
- Use cases in test `testfiles-regression/gh-666-tikz.lvt` are remained.\
  As now the manual examples use the new syntax only, maybe we can have a new test for old circle syntax.
- The changes to testfiles, two extra 0.0pt glues in `\showbox` output for each adaption of new circle syntax, are resulted from two uncommented newline spaces in the new circle syntax implementation + `\nullfont` applied inside `{tikzpictire}` env.\
  I plan to address most of those newline spaces and update testfiles in one go in the future.

**Checklist**

Please [signoff your commits][git-s] to explicitly state your agreement to the [Developer Certificate of Origin][DCO]. If that is not possible you may check the boxes below instead:

- [x] Code changes are licensed under [GPLv2][GPL] + [LPPLv1.3c][LPPL]
- [x] Documentation changes are licensed under [FDLv1.2][FDL]

[git-s]: https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s
[DCO]: https://developercertificate.org
[GPL]: https://www.gnu.org/licenses/gpl-2.0.html
[LPPL]: https://www.latex-project.org/lppl/lppl-1-3c.txt
[FDL]: https://www.gnu.org/licenses/fdl-1.2.html

PR #1471, opened 2026-08-20
test: add test for tikz circle, old syntax

muzimuzhi:

- Add test for old `circle` syntax, as I mentioned in #1467
  > As now the manual examples use the new syntax only, maybe we can have a new test for old circle syntax.
- <s>Add a new driver `pgfsys-test.def` dedicated for log-based testing, as mentioned in [matrix chat](https://matrix.to/#/!NuxCISwYQJuyWwNsEI:matrix.org/$DMrUPfrnPIwW7Co3QaxhieVjglo19Vjf9XW9-JuD9eQ?via=matrix.org&via=t2bot.io&via=freiburg.social).\
  Coverage of `\pgfsys@xxx` commands can be increased when needed.</s>\
  Superseded by #1472, see also the [comment](https://github.com/pgf-tikz/pgf/pull/1471#issuecomment-5352465311) below.

**Checklist**

### DIFF
--- a/testfiles-regression/tikz-circle-old-syntax.lvt
+++ b/testfiles-regression/tikz-circle-old-syntax.lvt
@@ -1,0 +1,35 @@
+% -*- mode: latex -*-
+% vim: ft=tex
+\documentclass{minimal}
+\input{regression-test}
+
+\RequirePackage{tikz}
+
+\begin{document}
+
+\START
+
+\BEGINBOXTEST{circle, old syntax}
+\tikzpicture
+  \draw (0,0) circle (1);
+  \draw (3,0) circle (10pt) circle (20pt);
+  \draw (6,0) circle (1cm) +(3,0) circle (2cm);
+\endtikzpicture
+\ENDBOXTEST
+
+\BEGINBOXTEST{ellipse, old syntax}
+\tikzpicture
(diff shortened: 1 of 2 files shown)


hmenke:
Instead of adding a dummy driver that we have to develop and maintain, I think we could also just run `testfiles-regression` for a single engine, e.g. `luatex`. Then we could also merge in `testfiles-regression-luatex`.

I have to check whether the latest l3build is already in the Docker image, then we can also get rid of our custom `pgf-regression-test.tex`.

I'll look into this.

hmenke:
Looking at this again, I am intrigued, although I'm not a big fan of having an extra driver that we have to maintain. Do you think we could instead hook the softpath, dump it to the log, and then clear it, so that the driver never engages?

muzimuzhi:
> Do you think we could instead hook the softpath, dump it to the log, and then clear it, so that the driver never engages?

Yes it's also feasible. But then non-softpath things cannot be dumped, while a custom driver could/should dump everything, even those not shown by `\showbox`, for example fadings and patterns.

muzimuzhi:
Also for debugging purposes, it seems to me it should not replace the normal output done by the chosen driver, but only writing logs.

Under the title of this PR, for now I'll drop the custom driver and only add a new test.

